Issue #8, opened 2025-08-28
Output window always opens


klueis:
At every startup, this extension opens the output console window. Although it might be possible to prevent this by setting the global override below, this should not be preferred. Please add setting to disable opening of the output console.
```
"workbench.view.showQuietly": {
    "workbench.panel.output": true
}
```

EDIT: Seems the above trick is not reliable either, so a setting to control this behavior in the extension is much preferred.

eFAILution:
Fix available in pre-release 0.5.x

PR #9, opened 2025-08-29
fix: prevent output window from automatically opening

eFAILution:
- Prevent output window from opening automatically during extension activation and debugging
- Introduce `autoShowOutput` setting to control output channel display
- Replace console.log with Logger calls for better logging management


eFAILution:
This PR closes #8 and will be included in pre-release 0.5.x (following VS Code marketplace convention: odd = pre-release, even = stable). The next stable release will be 0.6.0.

PR #125, opened 2026-05-29
feat: recognise non-canonical GitLab CI file names via additionalFileGlobs

### LINKED ISSUE #121: feat: recognise non-canonical GitLab CI file names via additionalFileGlobs
### Problem statement

GitLab CI lets you split a pipeline across many files by referencing them from `include:`, and those included files can be named anything and live anywhere in the repo — `ci/build.yml`, `pipelines/deploy.yaml`, `.gitlab/ci/lint.yml`, and so on. Today every provider (hover, completion, validation, code actions) decides "is this a GitLab CI file?" via the same four-way check, duplicated inline in each provider: `fileName.endsWith('.gitlab-ci.yml')`, `fileName.endsWith('.gitlab-ci.yaml')`, `fileName.includes('gitlab-ci')`, or `languageId === 'gitlab-ci'`. Anything else silently bails out. The substring check picks up incidental matches like `my-gitlab-ci-helper.yml`, but it doesn't help with the common case of files named after their purpose (`build.yml`, `deploy.yaml`) — and there's no way for a user to extend it.

The result is that authors who break their pipelines into multiple included files lose the extension's value in exactly those files — no component hover docs, no input completion, no include validation — even though the files contain the same syntax and reference the same components. There's no way for a user to tell the extension "this file is also a GitLab CI file".

### Proposed solution

Add a user-configurable list of glob patterns that augments the built-in detection, and consult it from every provider via a single helper:

1. New setting `gitlabComponentHelper.additionalFileGlobs` (array of GlobPattern strings, default `[]`). The user appends patterns for their nested/included CI files; the setting is *additive* — built-in defaults always apply.
2. Built-in defaults baked into the extension: `**/.gitlab-ci.yml`, `**/.gitlab-ci.yaml`, `**/.gitlab/**/*.yml`, `**/.gitlab/**/*.yaml`. The `.gitlab/` directory is GitLab's own convention for nested config, so it ships matched out of the box.
3. A single `isGitLabCIFile(document)` helper used by `HoverProvider`, `CompletionProvider`, and `ValidationProvider`, replacing the three near-duplicate inline checks. It returns true if either the language ID is `gitlab-ci` or the document matches any default-or-additional glob (via `vscode.languages.match`).
4. Activation events widened so the extension also wakes up on `workspaceContains:**/.gitlab/**/*.{yml,yaml}`.
5. Changes to `additionalFileGlobs` take effect immediately — a `workspace.onDidChangeConfiguration` listener re-validates open documents without requiring a window reload.

### Example .gitlab-ci.yml usage

```yaml
# .gitlab-ci.yml
include:
  - local: '.gitlab/ci/build.yml'         # picked up by default
  - local: 'ci/lint.yml'                  # needs additionalFileGlobs entry
  - local: 'pipelines/deploy.yaml'        # needs additionalFileGlobs entry

# settings.json
{
  "gitlabComponentHelper.additionalFileGlobs": [
    "**/ci/*.yml",
    "**/pipelines/**/*.yaml"
  ]
}

# .gitlab/ci/build.yml  — providers active automatically (default glob)
include:
  - component: $CI_SERVER_FQDN/components/opentofu/full-pipeline@2.9.0
    inputs:
      terraform_version: "1.5.0"

# ci/lint.yml  — providers active because of the user's additional glob
include:
  - component: $CI_SERVER_FQDN/components/sast/scan@1.2.0
    inputs:
      severity: high
```

### Command/UI impact

- New command: none
- Setting changes:
  - New: `gitlabComponentHelper.additionalFileGlobs` (array of GlobPattern strings, default `[]`, merged with built-in defaults)
  - Activation events widened to include `workspaceContains:**/.gitlab/**/*.yml` and `workspaceContains:**/.gitlab/**/*.yaml`
  - Editor context-menu `when` clause broadened from `resourceFilename =~ /gitlab-ci\.yml$/` to `resourceLangId == yaml || resourceLangId == gitlab-ci` so "GitLab CI: Browse Components" appears on any YAML file
- Browser/hover/completion changes: hover, completion, validation, and quick-fix code actions now activate on any document whose path matches a default or additional glob — no behaviour change for canonical `.gitlab-ci.yml` files
- Validation: the same provider runs on the additional files; existing `include:`/`component:`/inputs validation applies unchanged

### Acceptance criteria

- [ ] `.gitlab-ci.yml` and `.gitlab-ci.yaml` continue to work exactly as before (no regression in baseline behaviour)
- [ ] A file at `.gitlab/ci/build.yml` containing a `component:` include receives hover, completion, and validation without any user configuration
- [ ] Adding `"**/ci/*.yml"` to `gitlabComponentHelper.additionalFileGlobs` activates providers on `ci/build.yml` immediately, without a window reload
- [ ] Removing the additional glob deactivates providers on `ci/build.yml` while leaving `.gitlab/ci/build.yml` active
- [ ] A single helper (`src/utils/gitlabCiFileMatcher.ts`) is the only place that decides "is this a GitLab CI file?" — no duplicated `endsWith('.gitlab-ci.yml')` checks remain in the providers
- [ ] Unit test in `tests/unit/gitlab-ci-file-matcher.test.js` covers canonical paths, `.gitlab/` defaults, non-matching paths, language-ID override, and additional-glob behaviour
- [ ] README documents the new setting with a short example
- [ ] `npm run check-types`, `npm run lint`, and `npm test` all pass

### Priority

Important

### Alternatives considered

- **Content sniffing**: treat every YAML file as a candidate and inspect content for `include:` / `component:` / `stages:` keys. Zero configuration, but runs on every YAML open and would mis-identify unrelated files that happen to contain those keys. Rejected in favour of explicit globs.
- **Replace defaults instead of augmenting them**: a `fileGlobs` setting that fully overrides the canonical list. Simpler implementation but easy to footgun — a user who only wants to add one path glob ends up silently disabling the canonical defaults. Rejected in favour of additive semantics.

### Target GitLab instance(s)

both

### Pre-submit checks

- [x] I searched existing feature requests and did not find a duplicate.
- [x] I included concrete examples and acceptance criteria.

X-Guardian:
## Summary
- Adds `gitlabComponentHelper.additionalFileGlobs` so hover, completion, validation, and code actions activate on non-canonical GitLab CI file names (nested/included configs like `ci/build.yml`, `pipelines/deploy.yaml`). User-supplied globs are auto-normalised: an unanchored entry like `.gitlab2/**/*.yml` is treated as `**/.gitlab2/**/*.yml` so it matches at any directory depth.
- Ships built-in defaults that cover `**/.gitlab-ci.yml`, `**/.gitlab-ci.yaml`, and anything under `**/.gitlab/**/*.{yml,yaml}` — the conventional location for nested config — so `.gitlab/`-style layouts work out of the box.
- Consolidates the three near-duplicate `isGitLabCIFile` checks (hover, completion, validation) behind a single helper at `src/utils/gitlabCiFileMatcher.ts`. Globs are authoritative; `gitlab-ci` and `shellscript` documents are always-in-scope so user-defined language associations (`files.associations`) keep working. Helper memoises the combined glob list and invalidates on config change.
- Editor context-menu now gated on a `gitlabComponentHelper.isCiFile` context key that mirrors the matcher — set on activation, active-editor change, and config change — so "Browse Components" shows on exactly the documents the providers activate on, globs included.
- Link related issue(s): Closes #121

## Change Type
- [x] feat
- [ ] fix
- [x] refactor
- [ ] docs
- [ ] test
- [ ] chore

### User-facing impact
- Authors who split pipelines into multiple included files now get hover docs, completion, and include validation in those files too. The `.gitlab/` directory case requires no configuration; other layouts add a glob to `gitlabComponentHelper.additionalFileGlobs`.
- Changes to `additionalFileGlobs` take effect immediately — `workspace.onDidChangeConfiguration` re-validates open documents without a window reload.
- The legacy `fileName.includes('gitlab-ci')` substring escape hatch is gone. Files that previously activated only because their name contained the substring `gitlab-ci` (e.g. `my-gitlab-ci-helper.yml`) now need either a matching glob or the `gitlab-ci` language association. This is intentional but is the one observable narrowing — call out for the changelog.
- Validation tightening: validation no longer fires on every plain YAML document; it now requires a glob match or one of the always-in-scope language IDs (`gitlab-ci`, `shellscript`). Removes false-positive runs on unrelated YAML (Compose, k8s, etc.). shellscript documents remain always-in-scope so users who associate CI shell snippets with the shellscript language keep working.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both

### Affected areas
- [ ] Component Browser
- [x] Hover provider
- [x] Completion provider
- [x] Validation provider
- [ ] Docs only

### Local checks
- [x] `npm run compile`
- [ ] Manual verification in VS Code Extension Host

### Manual test notes
- Canonical baseline: open `.gitlab-ci.yml` with a `component:` include → hover, completion, validation all behave as before.
- `.gitlab/` default: create `.gitlab/ci/build.yml` containing
  ```yaml
  include:
    - component: $CI_SERVER_FQDN/components/opentofu/full-pipeline@2.9.0
      inputs:
        terraform_version: "1.5.0"
  ```
  → providers activate without any user configuration.
- User-added glob: with `ci/build.yml` (same content) open, add `"gitlabComponentHelper.additionalFileGlobs": ["**/ci/*.yml"]` → hover/completion/validation light up immediately, no reload.
- Remove the glob → providers go silent on `ci/build.yml` while remaining active on `.gitlab/ci/build.yml`.
- Unanchored user glob: with `repo/.gitlab2/foo/bar.yml` open, add `"gitlabComponentHelper.additionalFileGlobs": [".gitlab2/**/*.yml"]` (no leading `**/`) → providers activate. The setting is auto-normalised to match at any depth.
- Context menu: right-click in `.gitlab-ci.yml`, `.gitlab/ci/build.yml`, and any file matched via a custom glob → "GitLab CI: Browse Components" appears. Right-click in `docker-compose.yml` → it does not.
- Negative case: open a `docker-compose.yml` (plain YAML, no matching glob) → matcher returns false; no diagnostics from this extension.

## Breaking Changes
- [x] No breaking changes

No formal breaking change, but two observable narrowings worth noting:
1. The `fileName.includes('gitlab-ci')` substring escape hatch is removed. Affected users add the file's path to `additionalFileGlobs` or rely on a `gitlab-ci` language association.
2. Validation no longer fires on every plain YAML document. This is desirable (no more diagnostics on unrelated YAML like Compose or k8s manifests), and any user who genuinely had a non-canonical file working before via the substring path can opt in with `additionalFileGlobs`. shellscript documents remain always-in-scope; the narrowing is YAML-only.
3. Editor context menu now hidden on plain YAML files. Previously the `when` clause matched `gitlab-ci\.yml$`, so the menu showed on `.gitlab-ci.yml`-style files only — but in practice users opened other YAML and saw the same regex match. Now the menu uses a context key (`gitlabComponentHelper.isCiFile`) sourced from the matcher, so it appears on exactly the documents the providers activate on.

- n/a — no UI surface changes. Hover/completion/validation appearance is unchanged; only the set of files they activate on grows.

## Risk and Rollback
- Main risks:
  - A user relying on the removed substring escape hatch sees providers go quiet in files they expected to work. Mitigation: `additionalFileGlobs` covers every such case and is documented in the README.
  - VS Code's `GlobPattern` semantics differ subtly from minimatch in edge cases; the matcher delegates to `vscode.languages.match` to use the platform's own matcher, with the unit test mirroring expected behaviour via minimatch.
- Rollback strategy: revert this commit (single commit on the branch). Settings written by users to `additionalFileGlobs` are harmless after rollback — the old code ignores unknown settings.

## Release Notes Draft
- feat: recognise non-canonical GitLab CI file names via `additionalFileGlobs` (and out-of-the-box `**/.gitlab/**/*.{yml,yaml}`), with a single consolidated file-type matcher across providers and context menu.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,19 @@ Add your favorite sources in VS Code settings:
 ]
 ```
 
+### 📁 Recognising non-canonical CI files
+
+Out of the box the extension activates on `.gitlab-ci.yml`, `.gitlab-ci.yaml`, and anything under a `.gitlab/` directory. If your project keeps included CI configs elsewhere, add their globs to `gitlabComponentHelper.additionalFileGlobs` — these are merged with the built-in defaults:
+
+```jsonc
+"gitlabComponentHelper.additionalFileGlobs": [
+  "**/ci/*.yml",
+  "**/pipelines/**/*.yaml"
+]
+```
+
+Patterns use VS Code's GlobPattern syntax.
+
 ---
 
 ## 🗂️ Component Discovery
--- a/package.json
+++ b/package.json
@@ -17,8 +17,10 @@
   "activationEvents": [
     "onLanguage:yaml",
     "onLanguage:shellscript",
-    "workspaceContains:.gitlab-ci.yml",
-    "workspaceContains:.gitlab-ci.yaml"
+    "workspaceContains:**/.gitlab-ci.yml",
+    "workspaceContains:**/.gitlab-ci.yaml",
+    "workspaceContains:**/.gitlab/**/*.yml",
+    "workspaceContains:**/.gitlab/**/*.yaml"
   ],
   "contributes": {
     "languages": [
@@ -236,6 +238,14 @@
             "template.yaml"
           ],
           "description": "Filenames recognised inside per-component subfolders, e.g. 'templates/foo/template.yml'."
+        },
+        "gitlabComponentHelper.additionalFileGlobs": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": [],
+          "description": "Extra glob patterns for files that should be treated as GitLab CI files, in addition to the built-in defaults (`**/.gitlab-ci.yml`, `**/.gitlab-ci.yaml`, `**/.gitlab/**/*.yml`, `**/.gitlab/**/*.yaml`). Use this for nested/included CI configs that live elsewhere, e.g. '**/ci/*.yml'. Patterns use VS Code GlobPattern syntax."
         }
       }
     },
@@ -280,7 +290,7 @@
     "menus": {
       "editor/context": [
         {
-          "when": "resourceFilename =~ /gitlab-ci\\.yml$/",
+          "when": "resourceLangId == yaml || resourceLangId == gitlab-ci",
           "command": "gitlab-component-helper.browseComponents",
           "group": "navigation"
         }
--- a/package.json
+++ b/package.json
@@ -245,7 +245,7 @@
             "type": "string"
           },
           "default": [],
-          "description": "Extra glob patterns for files that should be treated as GitLab CI files, in addition to the built-in defaults (`**/.gitlab-ci.yml`, `**/.gitlab-ci.yaml`, `**/.gitlab/**/*.yml`, `**/.gitlab/**/*.yaml`). Use this for nested/included CI configs that live elsewhere, e.g. '**/ci/*.yml'. Patterns use VS Code GlobPattern syntax."
+          "description": "Extra GitLab CI file globs, merged with the built-in defaults. Patterns match at any directory depth (e.g. 'ci/*.yml' is treated as '**/ci/*.yml')."
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -290,7 +290,7 @@
     "menus": {
       "editor/context": [
         {
-          "when": "resourceLangId == yaml || resourceLangId == gitlab-ci",
+          "when": "gitlabComponentHelper.isCiFile",
           "command": "gitlab-component-helper.browseComponents",
           "group": "navigation"
         }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -10,6 +10,9 @@ import { ValidationProvider } from './providers/validationProvider';
 import { parseYaml } from './utils/yamlParser';
 import { DetachedComponentTemplate } from './templates';
 import { getPerformanceMonitor } from './utils/performanceMonitor';
+import { isGitLabCIFile, invalidateFileGlobsCache } from './utils/gitlabCiFileMatcher';
+
+const CI_FILE_CONTEXT_KEY = 'gitlabComponentHelper.isCiFile';
 
 // Constants for timing delays
 const PANEL_FOCUS_DELAY_MS = 100;
@@ -552,13 +555,26 @@ export function activate(context: vscode.ExtensionContext) {
     // Initialize the validation provider
     const validationProvider = new ValidationProvider(context);
 
+    // Keep the `gitlabComponentHelper.isCiFile` context key in sync with `isGitLabCIFile` so the editor context menu
+    // shows the Browse Components command on exactly the same files the providers activate on.
+    const updateCiFileContext = (editor: vscode.TextEditor | undefined): void => {
+      const isCi = editor ? isGitLabCIFile(editor.document) : false;
+      vscode.commands.executeCommand('setContext', CI_FILE_CONTEXT_KEY, isCi);
+    };
+    updateCiFileContext(vscode.window.activeTextEditor);
+    context.subscriptions.push(
+      vscode.window.onDidChangeActiveTextEditor(updateCiFileContext),
+    );
+
     // Re-validate open documents when the user changes additional file globs so newly-matched files light up (or stop
-    // being treated as GitLab CI) without reloading the window.
+    // being treated as GitLab CI) without reloading the window. Also refresh the context key so the menu follows suit.
     context.subscriptions.push(
       vscode.workspace.onDidChangeConfiguration(e => {
         if (e.affectsConfiguration('gitlabComponentHelper.additionalFileGlobs')) {
           logger.info('[Extension] additionalFileGlobs setting changed, re-validating open documents', 'Extension');
+          invalidateFileGlobsCache();
           validationProvider.revalidateOpenDocuments();
+          updateCiFileContext(vscode.window.activeTextEditor);
         }
       })
     );
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -550,7 +550,18 @@ export function activate(context: vscode.ExtensionContext) {
     );
 
     // Initialize the validation provider
-    new ValidationProvider(context);
+    const validationProvider = new ValidationProvider(context);
+
+    // Re-validate open documents when the user changes additional file globs so newly-matched files light up (or stop
+    // being treated as GitLab CI) without reloading the window.
+    context.subscriptions.push(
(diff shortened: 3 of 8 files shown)


eFAILution (commented):
Thanks Simon, clean and well-scoped change, and folding the three duplicate checks into one matcher is a real improvement. A few things before merge, roughly in priority order.

**1. Context-menu `when` clause drops `shellscript` (please fix).** `package.json` changes the "Browse Components" `when` from `resourceFilename =~ /gitlab-ci\.yml$/` to `resourceLangId == yaml || resourceLangId == gitlab-ci`. That omits `shellscript`, so on CI files opened under the shellscript language the menu command disappears, which is exactly the case we keep shellscript support for. Please add `|| resourceLangId == shellscript`. Also note this now shows the command on every YAML file (compose, k8s, and so on) even though hover/completion/validation no longer activate on plain YAML without a glob match, so the menu is broader than the features behind it. Fine if intentional, just flagging.

**2. No test exercises the real helper (please address or defer explicitly).** `tests/unit/gitlab-ci-file-matcher.test.js` reimplements the logic with `minimatch` instead of calling `isGitLabCIFile`/`getConfiguredFileGlobs`, which use `vscode.languages.match`. The header says end-to-end coverage lives in `tests/extension-host`, but this PR adds nothing there. So `minimatch(..., {dot:true})` is standing in for VS Code's GlobPattern engine, and the two can diverge (dotfile and `**` handling). A real bug in the helper would slip through. An extension-host test that runs the actual helper would close this, or drop the comment if you are deferring.

**3. shellscript comment + description wording.** Covered in my earlier comment: keep the registration, but the "surface inside `script:` blocks" rationale is incorrect (the real reason is the shellscript language association), and the description line "validation no longer fires on every YAML/shellscript document" is not accurate since shellscript stays always in scope. The narrowing applies to plain `yaml` only.

**4. Code-action selector captures globs at construction (minor).** `validationProvider` builds `codeActionSelectors` once in the constructor, so runtime changes to `additionalFileGlobs` do not update it. In practice it is masked because `{ language: 'yaml' }` is also in the selector and catches nearly every newly-matched file, and actions are diagnostic-driven. The glob entries here are close to redundant. Either drop them or note the reload caveat.

**5. Quote style (nit).** `gitlabCiFileMatcher.ts` uses double quotes; the rest of the codebase uses single. ESLint has no quotes rule so it passes lint, but please align for consistency.

**6. `isGitLabCIFile` re-reads config and iterates globs on every call (nit).** Runs per hover, per completion, and per-keystroke in `onDidChangeTextDocument`. Cheap today, just avoidable churn with large glob lists. Fine to leave.

Correctness, security, and the activation-event changes all look good. Nothing blocking on correctness; the main asks are #1 and #3.

PR #126, opened 2026-05-30
chore(test): PoC — Mocha + TypeScript unit test importing real src

X-Guardian:
## Summary
- **Proof of concept.** Migrates a single test file (`tests/unit/url-parsing.test.js` → `.ts`) to a new Mocha + tsx runner that imports the real `UrlParser` from `src/services/component/urlParser.ts` — replacing the previous inlined-reimplementation pattern documented in `tests/README.md`.
- Adds just enough infrastructure to support the pattern (Mocha config, tsx loader, JUnit XML output in CI, `dorny/test-reporter@v3.0.0` for inline GitHub annotations + a PR check-run, sentinel-based skip in the legacy runner so the two co-exist). The other 10 unit-test files are untouched and continue running through `tests/run-tests.js` exactly as before.
- The Logger / outputChannel refactor is the one thing this PoC *required* of `src/` — both modules now defer `vscode` access until first call, so they're loadable from plain Node. Without that, `UrlParser` couldn't be imported into a unit test at all.

## Change Type
- [ ] feat
- [ ] fix
- [ ] refactor
- [ ] docs
- [x] test
- [ ] chore

### User-facing impact
- None. The extension binary is unchanged in behaviour.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (no change to either)

### Affected areas
- [ ] Component Browser
- [ ] Hover provider
- [ ] Completion provider
- [ ] Validation provider
- [ ] Docs only

(Touches `src/utils/logger.ts` and `src/utils/outputChannel.ts`, but only their bootstrap — log behaviour inside the extension host is unchanged.)

## What this PoC proves
- A unit test can import a real `src/` module and assert against its actual contract (not a copy-paste of the algorithm).
- Failures in CI surface as **inline annotations on the PR diff** (via the JUnit XML → `dorny/test-reporter` chain) plus a **check-run summary** at the top of the PR listing every test by name.
- The legacy runner and Mocha can run side-by-side during a phased migration; a `// @mocha` sentinel on the first line of a file tells the legacy runner to skip it.

## How CI reporting works now
- The "Run Mocha unit tests" step runs Mocha with the built-in `xunit` reporter (`GITHUB_ACTIONS=true` → `.mocharc.cjs` selects `xunit` and writes `junit.xml`). Step succeeds with exit 0; the raw log itself is quiet.
- The "Publish Mocha test report" step runs `dorny/test-reporter@v3.0.0` with `reporter: java-junit, path: junit.xml`. It parses the XML and emits a Check Run with full test list + inline annotations on any failures.
- `fail-on-error: true` on the reporter step preserves the existing semantics that a failing test fails the job.
- No third-party Mocha reporter packages added — `xunit` is built into Mocha itself.

## What's intentionally *not* in this PR
- The other 10 unit-test files. Each will be its own follow-up — most need the src-side function they want to test exposed (or extracted) before the inline copy can be deleted.
- The extension-host suite. `tests/extension-host/runTest.ts` has a pre-existing path bug (broken since `2c748e6`); not addressed here.
- c8/coverage reporting. Worth adding once more files have been migrated; premature on one file.

### Local checks
- [x] `npm run compile` — esbuild bundle still builds (`out/extension.js`)
- [x] `npm test` — legacy runner: 10/10 passing (the migrated file is now `.ts` and lives under Mocha)
- [x] `npm run test:mocha` — 8/8 against real `UrlParser`, spec reporter output in terminal
- [x] `GITHUB_ACTIONS=true npm run test:mocha` — writes `junit.xml`, valid JUnit XML (testsuite/testcase/failure elements)
- [x] `npm run check-types` — clean (now also type-checks unit tests via `tsconfig.tests.json`)
- [ ] Manual verification in VS Code Extension Host (logger lazy-init path: recommend a quick F5 sanity-check that the output channel still appears on `Show Output`)

### Manual test notes
- Forced a deliberate `assert.strictEqual` failure and confirmed the failure shows up correctly inside the `<failure>` element of the JUnit XML, including file path, line, and assertion message — exactly what `dorny/test-reporter` needs to produce a useful annotation.

## Breaking Changes
- [x] No breaking changes

- N/A — no UI changes.

## Risk and Rollback
- **Main risks:**
  - The Logger refactor preserves runtime behaviour exactly: the singleton is unchanged, the `vscode.workspace.onDidChangeConfiguration` subscription still happens synchronously in the constructor when `vscode` is resolvable, and the only new branch is a safe no-op when `vscode` is absent (plain Node, e.g. unit tests). Risk is minimal.
  - `dorny/test-reporter@v3.0.0` requires Node 24 on the GHA runner (separate from project Node). `ubuntu-latest` already provides Node 24, so this works out of the box.
- **Rollback strategy:** Single PR — revert. No data migration, no settings shape change, no CI secrets touched, `junit.xml` is gitignored.

## Release Notes Draft
- Internal: PoC migrating unit tests to Mocha + TypeScript with real-`src/` imports; one file converted, infrastructure in place for the rest.

## Checklist
- [x] Added/updated docs for behavior or settings changes (none required)
- [x] Added/updated tests for new behavior (the new test framework *is* the change)


eFAILution (commented):
Thanks Simon. This is mechanically clean and the direction (real-`src` imports from unit tests) is the right one. A few notes.

The lazy-`vscode` refactor of `logger.ts`/`outputChannel.ts` is behavior-preserving: the config subscription is still in the constructor, the singleton is unchanged, and the Node fallback is a safe no-op. Worth tweaking the PR description though, it says the subscription is deferred to the first log call, but it isn't (still in the constructor), so the risk it describes doesn't actually exist.

One thing to change before merge: `dorny/test-reporter@v3.0.0` in `ci.yml` is a floating tag. Could you SHA-pin it (with a trailing `# v3.0.0` comment)? It's a new third-party action and we want those pinned. The existing actions in that workflow are also floating tags, but no need to touch those here, just the new one.

On necessity: fine to land as a PoC. Just flagging that the payoff is mostly in the follow-ups, it moves one test and deliberately leaves the extension-host suite (broken since 2c748e6) alone, so it doesn't move the coverage needle on its own yet. No objection to merging the scaffolding.

Process note: this should target `beta` rather than `main`. Features and infra land on `beta` and roll into the release through the `beta -> main` integration PR; PRs into `main` are treated as release events. Same retarget as #125.

Issue #128, opened 2026-05-30
feat: support local includes as first-class components


X-Guardian:
### Problem statement

The extension only recognises catalog-style `include: - component:` entries. Real-world `.gitlab-ci.yml` files frequently use `include: - local:` to pull in templates from the same repository, and those local templates often define a `spec.inputs` block — the same schema catalog components use. In those files the extension provides no hover, completion, validation, or browser support, so users editing `inputs:` under a `- local:` include get none of the assistance they would get for a `- component:` include of an equivalent template.

Example — a project's `.gitlab-ci.yml` uses:

```yaml
include:
    - local: "gitlab/templates/nx-test/template.yml"
      inputs:
        job_name: test-nightly
        job_type: nightly
```

…and the referenced local template declares a full `spec.inputs` block (`job_name`, `job_type` with enum options, `git_depth`, `debug`, etc.). The user gets zero IDE assistance for that `inputs:` map even though every piece of information needed is sitting in a file on disk.

### Proposed solution

Extend [src/providers/componentDetector.ts](src/providers/componentDetector.ts) and the surrounding cache/resolution layer so that `- local:` includes are treated as a first-class component source:

1. **Detection** — recognise `local:\s*["']?(.+?)["']?$` on `include:` entries in addition to the existing `component:` regex at [componentDetector.ts:85](src/providers/componentDetector.ts#L85).
2. **Path resolution** — resolve the path relative to the repository root (GitLab's own rule for `local:`), via the VS Code Git extension, falling back to the workspace folder when no Git repo is detected. Glob patterns and `../` traversal are out of scope.
3. **Parsing** — read the target YAML, look for a top-level `spec.inputs` block, and synthesise a `Component` record with `name = <relative path>`, `source = "local"`, `inputs = <parsed inputs>`. No HTTP fetch needed.
4. **Surface** — feed those synthesised components through the existing cache so hover, completion, validation, and the Component Browser all light up identically to catalog components.

Local components don't need a version (the file is the source of truth), so the existing `@version` handling can be skipped for this source type.

### Example .gitlab-ci.yml usage

```yaml
# Before — no IDE assistance on the `inputs:` block:
include:
    - local: "gitlab/templates/nx-test/template.yml"
      inputs:
        job_name: test-nightly
        job_type: nightly            # no completion for "affected" | "nightly" | "modified:nightly"
        git_depth: 0                 # no type/default hover

# After — full assistance:
[...]

PR #129, opened 2026-05-30
feat: support local includes as first-class components

X-Guardian:
## Summary
- Treats `include: - local: "<path>"` entries as first-class components alongside `include: - component:`. Hover, completion, and validation now activate on the `inputs:` block under a local include exactly as they do for catalog components — the same provider surfaces, the same diagnostics, just sourced from a workspace file instead of the GitLab CI Catalog.
- New `src/providers/localComponentResolver.ts` does the lookup: extracts the path from the include line, resolves it relative to the workspace folder, parses the target YAML with `js-yaml`'s `loadAll`, and maps the first document's `spec.inputs` block onto the existing `Component` shape. No HTTP calls, no cache writes — local files are the source of truth and re-resolved on demand.
- Wires the resolver into `detectIncludeComponent` (componentDetector), the input-walker loops in hoverProvider and completionProvider, and a new `validateLocalInclude` branch in validationProvider. Dispatch is keyed off `include.local && !include.component`, not URL shape — no shape-sniffing in `findComponentInCache`.
- Drive-by fix: `tests/extension-host/runTest.ts` walked two `..` levels from `__dirname`, but `tsconfig.test.json`'s `rootDir` flattens the compiled `runTest.js` to `out-test/runTest.js` (depth 1), so the workspace argument resolved outside the repo. The existing host tests passed by accident — they opened files via absolute URI and never touched workspaceFolders. The new local-include tests *do* depend on workspaceFolders, which surfaced the bug. One-line fix: `path.resolve(__dirname, '..')`. Dead since 2c748e6.
- Link related issue(s): Closes #128

## Change Type
- [x] feat
- [ ] fix
- [ ] refactor
- [ ] docs
- [ ] test
- [ ] chore

### User-facing impact
- Editing `inputs:` under a `- local:` include now gets the same assistance as `- component:` includes: hover on the include line shows the parameters table, hover on an input parameter shows that input's description/type/default, completion suggests missing inputs, and validation flags unknown inputs and missing required ones.
- Local includes that don't define `spec.inputs` are silently ignored (single debug log, no toast, no diagnostic). Missing local files produce a single warning diagnostic (`local-include-not-found`) rather than a stream of unknown-input warnings.
- No new settings, no new commands. Behaviour change is purely additive: files that didn't have local includes before continue to behave identically.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both

The feature is filesystem-only — no GitLab API calls — so it's inherently neutral to the instance the user targets.

### Affected areas
- [ ] Component Browser
- [x] Hover provider
- [x] Completion provider
- [x] Validation provider
- [ ] Docs only

### Local checks
- [x] `npm run compile`

### Manual test notes
Tested in the Extension Host against a real-world `.gitlab-ci.yml` that uses:

```yaml
include:
    - local: "gitlab/templates/nx-test/template.yml"
      inputs:
        job_name: test-nightly
        job_type: nightly
```

- Hover on the `- local:` line → parameters table populated from the template's `spec.inputs`.
- Hover on `job_type:` (an input parameter under the include) → "Input Parameter: job_type" panel with the description and default from the template.
- Add an unknown input (`bogus_input: oops`) → yellow `unknown-input` underline; "Unknown input 'bogus_input' for local include '<filename>'".
- Rename the local file → red `local-include-not-found` underline on the include line; no spurious unknown-input warnings for the inputs that were valid before the rename.
- File a `- local:` that points at a YAML without a `spec.inputs` block (e.g. a `.cache-ci.yml`) → no diagnostics, no toasts; single debug log line in the output channel.

### Automated tests
- **Unit** ([tests/unit/local-include-detection.test.js](tests/unit/local-include-detection.test.js)) — 11 cases characterising the `extractLocalIncludePath` regex: quoted/unquoted/single-quoted paths, leading slash, list-dash present or absent, trailing whitespace, and negatives for `component:`/`project:`/`remote:`, empty lines, and comments containing `local:`.
- **Extension-host** ([tests/extension-host/suite/localInclude.test.ts](tests/extension-host/suite/localInclude.test.ts)) — four end-to-end scenarios driven through the real provider stack against fixtures in [tests/fixtures/local-include/](tests/fixtures/local-include/): hover-on-include surfaces inputs, hover-on-input shows parameter docs, unknown input produces a diagnostic, missing local file produces a not-found diagnostic.

## Breaking Changes
- [x] No breaking changes

The change is purely additive — existing `- component:` includes go through the unchanged code path. Local-include detection short-circuits inside `detectIncludeComponent` and the validation loop before any catalog logic runs, so files without `- local:` entries see zero behavioural difference.

- n/a — the hover/completion/diagnostic UI is identical to the existing catalog-component flow. Only the set of include types that drive those surfaces has grown.

## Risk and Rollback
- Main risks:
  - **YAML walker fragility.** The hover/completion provider walkers find the include line by substring-matching `local:` against the document lines. A line containing the substring `local:` somewhere unrelated (e.g. in a value or comment) could in principle false-match. This is the same risk the existing `'component:'` walker has carried; the new code uses an `isLocal` flag derived from the parsed YAML node, so dispatch is correct even when the substring search picks a slightly off line.
  - **`js-yaml.loadAll` on large templates.** The resolver reads and parses the entire target file on every hover/completion/validation. For typical CI templates (single-digit KB) this is imperceptible. If users point local includes at multi-MB YAML files the parse cost could show up; mitigation is the existing 5-second parse cache in `src/utils/yamlParser.ts`, though the resolver currently calls `yaml.loadAll` directly to support multi-doc files — worth revisiting if a real complaint appears.
  - **No cache invalidation on local file save.** The resolver re-reads the local file on every detection call, so saves to the template are picked up immediately by hover/completion/validation without any explicit invalidation step. The cost is the per-call file read; the benefit is zero stale-state risk.
- Rollback strategy: revert the feature branch as a single PR. The new module (`localComponentResolver.ts`) is the only added file; the changes to existing providers are small, self-contained branches gated on `include.local && !include.component` that produce no effect when no local includes are present.

## Release Notes Draft
- feat: support `include: - local:` entries as first-class components — hover, completion, and validation now surface `spec.inputs` from local templates.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@
 - **Smart Completion**: Context-aware suggestions for components and versions as you type
 - **Hover Docs**: See full documentation and parameter hints instantly
 - **Input Validation**: Real-time validation of component inputs with intelligent Quick Fix suggestions
+- **Local Includes**: Same hover, completion, and validation for `include: - local:` entries that declare a `spec.inputs` block
 - **Version/Tag Picker**: Always use the right version—no more guessing
 - **Variable Expansion**: Full support for GitLab CI/CD variables in URLs and parameters
 - **Lightning Fast**: Caching, batch API calls, and performance optimizations for huge catalogs
@@ -61,6 +62,18 @@ include:
       workspace: "default"
       apply: true
 ```
+
+Local templates work too — point a `- local:` entry at any workspace YAML that declares a `spec.inputs` block and you get the same hover, completion, and validation as a catalog component:
+
+```yaml
+include:
+  - local: "gitlab/templates/nx-test.yml"
+    inputs:
+      job_name: test-nightly
+      job_type: nightly
+```
+
+The target file is resolved relative to the workspace root and re-read on demand, so edits to the template are picked up immediately.
 
 ### 🎬 Adding Component Inputs
 ![insertInputs](https://github.com/user-attachments/assets/098f4eaf-3c4a-45a8-9caf-9a1351730b93)
--- a/README.md
+++ b/README.md
@@ -73,8 +73,6 @@ include:
       job_type: nightly
 ```
 
-The target file is resolved relative to the workspace root and re-read on demand, so edits to the template are picked up immediately.
-
 ### 🎬 Adding Component Inputs
 ![insertInputs](https://github.com/user-attachments/assets/098f4eaf-3c4a-45a8-9caf-9a1351730b93)
 
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -6,6 +6,7 @@ import { getComponentCacheManager } from '../services/cache/componentCacheManage
 import { getVariableCompletions, GITLAB_PREDEFINED_VARIABLES, containsGitLabVariables, expandComponentUrl } from '../utils/gitlabVariables';
 import { Logger } from '../utils/logger';
 import { isGitLabCIFile } from '../utils/gitlabCiFileMatcher';
+import { resolveLocalComponent } from './localComponentResolver';
 
 export class CompletionProvider implements vscode.CompletionItemProvider {
   private logger = Logger.getInstance();
@@ -394,15 +395,15 @@ export class CompletionProvider implements vscode.CompletionItemProvider {
 
       // Find which component's inputs section we're in
       for (const include of includes) {
-        if (!include.component) continue;
-
-        // Find the component in the document text
-        const componentUrl = include.component;
+        const isLocal = !!include.local && !include.component;
+        const componentUrl = include.component ?? include.local;
+        if (!componentUrl) continue;
+        const includeKey = isLocal ? 'local:' : 'component:';
 
         // Look for this component's position in the file
         let componentLineIndex = -1;
         for (let i = 0; i < lines.length; i++) {
-          if (lines[i].includes('component:') && lines[i].includes(componentUrl)) {
+          if (lines[i].includes(includeKey) && lines[i].includes(componentUrl)) {
             componentLineIndex = i;
             break;
           }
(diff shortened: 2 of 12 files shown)


eFAILution (commented):
Thanks Simon, this is a genuinely useful feature and the in-flight, no-cache approach is exactly right (it sidesteps the whole versioning/cache-identity problem). A few things to work through before merge, mostly around matching GitLab's real `local:` semantics.

Three correctness points:
1. Path resolution anchors on the VS Code workspace folder (`localComponentResolver.ts:167-185`), but GitLab resolves `local:` relative to the repository root. Those differ for a monorepo subdir, a nested repo, or a multi-root workspace, and resolution will then point at the wrong file or none. The README states the workspace-folder rule as if it were GitLab's, worth correcting too.
2. No glob support (`:153-156`). `local: "configs/*.yml"` is read as a literal filename, so it throws and produces a spurious `local-include-not-found`. Globs were in the #128 acceptance criteria. If you'd rather defer them that's reasonable, but then it should skip (not warn) on a path containing glob characters.
3. No `../` guard (`:178-185`). `local: "../../x"` resolves and gets read. Low blast radius since it only parses YAML, but GitLab forbids `../`, so worth rejecting.

Smaller notes: it reads from disk, so unsaved edits to an open template aren't reflected (the README's "picked up immediately" is true only for saved edits; reading the open TextDocument when one exists would close that gap), and it parses via `yaml.loadAll` directly rather than the 5s parse cache (bounded by the 300ms validation debounce, so fine for now). The `local:` walker false-match you flagged is correctly handled by the `isLocal` flag.

Nice catch on the `runTest.ts` fix, that one line actually un-breaks the extension-host runner (dead since 2c748e6), so your new tests and the previously-dead ones now run. Could you add a missing-required-input case? The current fixtures only exercise unknown-input and missing-file.

On process: this implements #128, which we're still actively scoping (I just left design questions on the issue around repo-root, globs, and whether a missing file should be a diagnostic). This PR effectively answers those, and a couple of the answers are the mismatches above, so let's reconcile the two. Could you link #128 in the description? And like the others, this should target `beta`, not `main`.

PR #131, opened 2026-05-30
fix: component: URLs and Component Browser "Template File" links

### LINKED ISSUE #130: bug: component: URLs and Component Browser "Template File" links are incorrect
### Bug summary

The extension doesn't produce a link for `component:` URLs in `.gitlab-ci.yml`. VS Code's built-in URL detection underlines the literal `https://...` form, but cmd-clicking that 404s — GitLab's `component:` syntax (`https://<instance>/<project-path>/<component-name>@<ref>`) is a catalog identifier shaped like a URL, not a real navigable web URL. The "Template File" link in the Component Browser detail view is also wrong: a hardcoded heuristic always produces `templates/<name>.yml`, which 404s for repos using the directory layout (`templates/<name>/template.yml`). The hover card's "Source" link has the same problem — it points at the project root (`https://<instance>/<path>`), not at the template itself, so it doesn't take you to the file you'd expect from clicking it. Finally, the "Component not in cache" fallback hover renders the raw `component:` URL as a clickable link in its body, which also 404s for the same catalog-identifier reason.

A secondary issue: when the `component:` URL uses GitLab variables like `$CI_SERVER_FQDN/...`, even the (wrong) default underline disappears — VS Code's built-in URL detection only recognises `https://` strings, so without an extension-provided link there's nothing to click at all.

### Steps to reproduce

1. Open a `.gitlab-ci.yml` with both URL forms:

   ```yaml
   include:
     - component: $CI_SERVER_FQDN/my-group/shared-ci/my-component@v1.2.0
     - component: https://gitlab.com/my-group/shared-ci/my-component@v1.2.0
   ```

2. Observe the first line is plain text — no underline.
3. Observe the second line is underlined by VS Code's built-in URL detection.
4. Cmd-click the second line — lands on `https://gitlab.com/my-group/shared-ci/my-component@v1.2.0`, which 404s. The project is `my-group/shared-ci`; `my-component` is a catalog identifier inside it, not a navigable subpath, so the URL doesn't resolve on GitLab's web UI even ignoring the `@<ref>` suffix.
5. Open the Component Browser, view the component's detail. Observe the "Template File" row links to `https://gitlab.com/.../-/blob/v1.2.0/templates/my-component.yml`, which 404s in repos where the template lives at `templates/my-component/template.yml` (directory form).
6. Hover over a `component:` URL. Observe the "Source" row links to `https://gitlab.com/my-group/shared-ci` — the project root, not the template — so clicking it doesn't take you to the file you'd expect.
7. Hover over a `component:` URL for a component not yet in the cache (e.g. before the first refresh, or a project not in `componentSources`). Observe the fallback hover card includes a clickable `URL: https://gitlab.com/.../my-component@v1.2.0` line that 404s when clicked.

### Expected behavior

- Both URL forms are underlined and cmd-clickable.
- Cmd-click navigates to the **actual template location** on the source GitLab project at the pinned ref. For directory-form templates (`templates/<name>/template.yml`) the link should point at the directory; for single-file form (`templates/<name>.yml`) it should point at the file.
- The "Template File" link in the Component Browser and the "Source" link in the hover card both use the same logic, so all three surfaces agree.
- The "Component not in cache" fallback hover should not present the malformed catalog identifier as a clickable URL — the user already knows what they typed, and clicking it just 404s.

### Actual behavior

- Variable-form URLs are not detected as links at all.
- Literal-form URLs get VS Code's default URL underline, but cmd-clicking navigates to a 404. GitLab's component URL shape is an identifier, not a real web URL, so opening it verbatim doesn't resolve to anything — neither with the `@<ref>` suffix nor without it.
- The Component Browser's "Template File" link is hardcoded to single-file form and 404s for any project using the directory layout.
- The hover card's "Source" link points at the project root, not at the template — clicking it lands you on the project's main page instead of the file.
- The "Component not in cache" fallback hover renders the raw `component:` URL as a clickable link in its body, which 404s when clicked.

Component URL: `$CI_SERVER_FQDN/my-group/shared-ci/my-component@v1.2.0`
GitLab instance: `gitlab.com`
Project path: `my-group/shared-ci`
Version/reference: `v1.2.0`
Source type: project
Repo layout: directory form (`templates/my-component/template.yml` exists; no single-file `.yml` at `templates/my-component.yml`).

```yaml
include:
  - component: $CI_SERVER_FQDN/my-group/shared-ci/my-component@v1.2.0
    inputs:
      job_name: "TODO: set value1"
  - component: https://gitlab.com/my-group/shared-ci/my-component@v1.2.0
    inputs:
      job_name: "TODO: set value2"
```

### Extension logs

No runtime error — the extension never produced a `DocumentLink` for `component:` URLs at all, so VS Code falls back to its built-in URL detection which doesn't understand GitLab's `path@ref` syntax. The Component Browser's "Template File" URL is computed by a hardcoded heuristic in `componentBrowserProvider.buildTemplateFileUrl()` that always assumes single-file form.

### Affected area

- [ ] Component Browser
- [ ] Hover provider
- [ ] Completion provider
- [ ] Validation provider
- [ ] Cache behavior
- [ ] GitLab API integration
- [ ] Token/authentication
- [x] Other

(Other: editor `DocumentLinkProvider` integration + Component Browser "Template File" link rendering.)

### Extension version

0.10.4

### VS Code version

1.109.0 or later

### Node.js version

22.x

### Operating system

macOS (development); reproduces everywhere — bug is in the extension's rendering logic, not platform-specific.

### Pre-submit checks

X-Guardian:
## Summary
- Registers a `DocumentLinkProvider` for `component:` URLs in GitLab CI files so the URL is underlined and cmd-clickable — including when the URL uses GitLab variables like `$CI_SERVER_FQDN`, which VS Code's built-in URL detection ignores.
- Routes link generation through a shared `templateFileUrl` helper that uses the *resolved* template path recorded by the fetcher. The URL shape depends on the on-repo layout: single-file form (`templates/<name>.yml`) → `/-/blob/<ref>/<path>`; directory form (`templates/<name>/template.yml`) → `/-/tree/<ref>/templates/<name>`.
- Extends the cache schema with `templatePath` and persists it from every fetch path (catalog discovery, per-URL dynamic fetch, and the catalog branch of `fetchComponentMetadata`). Bumps cache version to `1.1.0` so caches written by older versions are discarded on load and repopulated on next refresh — preferred over field-level backfill heuristics because a single schema-version check is much less surface to maintain as fields are added over time.
- Replaces the Component Browser's hardcoded single-file `templates/<name>.yml` heuristic and the hover card's "Source" link (which previously pointed at the project root) with the same shared helper, so the inline editor link, the Component Browser's "Template File" link, and the hover's "Source" link all agree.
- Drops the misleading `URL: <component-url>` line from the "Component not in cache" fallback hover — the URL was rendered as a clickable link that 404'd, and it duplicated information the user just typed.
- Link related issue(s): Fixes #130

## Change Type
- [ ] feat
- [x] fix
- [ ] refactor
- [ ] docs
- [ ] test
- [ ] chore

### User-facing impact
- `component:` URLs are now underlined and cmd-clickable in `.gitlab-ci.yml` files, including the `$CI_SERVER_FQDN/...` variable form.
- Cmd-click navigates to the actual template location on GitLab — the directory for directory-form repos, the file for single-file repos — instead of VS Code's default behaviour of cmd-clicking the literal `component:` string, which is a catalog identifier (not a real web URL) and 404s on GitLab.
- Component Browser's "Template File" row and the hover card's "Source" link now both agree with the inline link.
- For components not in the cache, the fallback hover no longer shows a clickable (broken) URL line — just the failure reason and the suggestion to refresh.
- If a component's URL is for a project that isn't in the user's configured `componentSources`, no link is produced. The same applies in the brief window before the first cache refresh completes after extension activation.
- First run after upgrade: a one-time `Cache schema upgrade (1.0.0 -> 1.1.0)` line appears in the GitLab Component Helper output channel; the cache is discarded and repopulates on next refresh.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both

### Affected areas
- [x] Component Browser
- [x] Hover provider
- [ ] Completion provider
- [ ] Validation provider
- [x] GitLab API calls/auth/token storage
- [ ] Docs only

Other areas touched (not listed above):
- New `DocumentLinkProvider` registered against `language: 'yaml'` and `language: 'gitlab-ci'`.
- New shared helper `src/utils/templateFileUrl.ts`.
- Component Browser webview-side JS no longer duplicates the routing logic — the server precomputes the template-file URL and passes it through the message payload.
- Fetcher (`componentFetcher.ts`) always probes the template via `fetchTemplate` in the catalog branch of `fetchComponentMetadata` to learn the on-repo path — one extra HTTP call per catalog-hit metadata fetch, gated to that branch.

### Local checks
- [x] `npm run compile`

### Manual test notes
Tested in the Extension Development Host with a `.gitlab-ci.yml` containing both URL forms:

```yaml
include:
  - component: $CI_SERVER_FQDN/my-group/shared-ci/my-component@v1.2.0
    inputs:
      job_name: "TODO: set value1"
  - component: https://gitlab.com/my-group/shared-ci/my-component@v1.2.0
    inputs:
      job_name: "TODO: set value2"
```

- Both lines now show an underlined link.
- Cmd-click on either lands on `https://gitlab.com/my-group/shared-ci/-/tree/v1.2.0/templates/my-component` — the actual template directory, matching the repo's directory-form layout.
- Component Browser detail view's "Template File" link matches; hover card's "Source" link matches.
- Hovering over a component not in the cache shows the fallback hover with no URL line (previously a clickable, broken URL).
- Verified the schema-version invalidation: with an existing pre-upgrade cache on disk, reloading the extension logs `[ComponentCache] Cache schema upgrade (1.0.0 -> 1.1.0)`, discards the cache, repopulates via the next refresh, and links appear correctly. Direct cache resets behave the same way.

Unit tests: 6 cases in `tests/unit/component-link-target.test.js` covering single-file form, directory form (multiple project paths), absent version (default-branch fallback), non-default `templateRoots`, and refs containing slashes. Full suite passes (14/14).

## Breaking Changes
- [x] No breaking changes

The cache version bump invalidates existing on-disk caches, which then repopulate transparently on next refresh. No settings changes, no user action required.

## Risk and Rollback
- **Main risks:**
  - Cache invalidation is one-time and affects all users on first launch after the upgrade. If the refresh fails (network, auth), users may see "Component not in cache" messaging until repopulation succeeds — the same surface as today's behaviour when the cache is empty for any other reason, just affecting everyone at once. Chose this over field-level backfill because one schema-version check is far less code to maintain as the schema evolves.
  - Components from projects not listed in `componentSources` produce no link (intentional — no resolved path means no link). If anyone was relying on VS Code's broken default URL underline as a navigation aid for unconfigured projects, that goes away. Acceptable trade-off: the previous behaviour landed on 404s anyway, since GitLab component URLs aren't real navigable web URLs.
  - VS Code's built-in URL detection still fires on `https://...` substrings; extension-contributed links take precedence in the link picker so cmd-click hits the new provider's link, but the underline rendering may overlap visually in some themes.
- **Rollback strategy:** revert the commit. The cache version goes back to `1.0.0`; users who upgraded then rolled back see one more cache invalidation as the new version is treated as a downgrade. No data loss.

## Release Notes Draft
- Fix: `component:` URLs in GitLab CI files are now underlined and cmd-clickable, including when they use `$CI_SERVER_FQDN` or other GitLab variables. The link points at the actual template file or directory on the source GitLab project, matching the repo's layout.

## Checklist
- [ ] Added/updated docs for behavior or settings changes

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2,6 +2,7 @@ import { registerAddProjectTokenCommand, getComponentService } from './services/
 import * as vscode from 'vscode';
 import { HoverProvider } from './providers/hoverProvider';
 import { CompletionProvider } from './providers/completionProvider';
+import { ComponentDocumentLinkProvider } from './providers/documentLinkProvider';
 import { ComponentBrowserProvider } from './providers/componentBrowserProvider';
 import { detectIncludeComponent } from './providers/componentDetector';
 import { getComponentCacheManager, ComponentCacheManager } from './services/cache/componentCacheManager';
@@ -91,6 +92,19 @@ export function activate(context: vscode.ExtensionContext) {
         ],
         new CompletionProvider(),
         ':', ' ', '@'  // Add @ as a trigger character for version completions
+      )
+    );
+
+    // Register document link provider so `component:` URLs (including those using GitLab variables like
+    // $CI_SERVER_FQDN) become clickable and point at the GitLab project tree at the requested ref.
+    logger.debug('[Extension] Registering document link provider...', 'Extension');
+    context.subscriptions.push(
+      vscode.languages.registerDocumentLinkProvider(
+        [
+          { language: 'yaml' },
+          { language: 'gitlab-ci' }
+        ],
+        new ComponentDocumentLinkProvider()
       )
     );
 
--- a/src/providers/componentBrowserProvider.ts
+++ b/src/providers/componentBrowserProvider.ts
@@ -1196,8 +1196,7 @@ export class ComponentBrowserProvider {
           function viewDetailsById(componentName, version, projectId) {
             const componentData = window.componentVersionData[componentName];
             if (componentData && componentData[version]) {
-              // Send the raw component data; the server computes the template-file URL when it
-              // renders the details panel.
+              // Send the raw component data; the server computes the template-file URL when it renders the details panel.
               const component = {
                 ...componentData[version],
                 name: componentName,
--- a/src/providers/componentBrowserProvider.ts
+++ b/src/providers/componentBrowserProvider.ts
@@ -5,6 +5,7 @@ import { GitLabCatalogComponent, GitLabCatalogVariable } from '../types/gitlab-c
 import { Component, ComponentParameter } from './componentDetector';
 import { containsGitLabVariables, expandGitLabVariables } from '../utils/gitlabVariables';
 import { Logger } from '../utils/logger';
+import { templateFileUrlForResolved } from '../utils/templateFileUrl';
 
 // Constants for timing delays
 const EDITOR_ACTIVATION_DELAY_MS = 50;
@@ -510,10 +511,14 @@ export class ComponentBrowserProvider {
               selectedVersion
             );
             if (updatedComponent) {
-              // Send the updated component details to the webview
+              // Send the updated component details to the webview, with the template-file URL precomputed
+              // server-side so the webview never has to do its own URL routing.
               detailsPanel.webview.postMessage({
                 command: 'componentDetailsUpdated',
-                component: updatedComponent
+                component: {
+                  ...updatedComponent,
+                  templateFileUrl: this.buildTemplateFileUrl(updatedComponent),
+                }
               });
             } else {
               detailsPanel.webview.postMessage({
@@ -1191,11 +1196,12 @@ export class ComponentBrowserProvider {
           function viewDetailsById(componentName, version, projectId) {
             const componentData = window.componentVersionData[componentName];
             if (componentData && componentData[version]) {
+              // Send the raw component data; the server computes the template-file URL when it
+              // renders the details panel.
               const component = {
                 ...componentData[version],
                 name: componentName,
                 version: version,
-                templateFileUrl: buildTemplateFileUrl(componentData[version], componentName, version)
               };
               vscode.postMessage({ command: 'viewComponentDetails', component });
             }
@@ -1213,14 +1219,6 @@ export class ComponentBrowserProvider {
               };
               vscode.postMessage({ command: 'insertComponent', component });
             }
-          }
-
-          function buildTemplateFileUrl(versionData, componentName, version) {
-            if (!versionData || !versionData.sourcePath || !versionData.gitlabInstance) {
-              return '';
-            }
-            const ref = version || 'main';
-            return 'https://' + versionData.gitlabInstance + '/' + versionData.sourcePath + '/-/blob/' + ref + '/templates/' + componentName + '.yml';
           }
 
           function filterComponents() {
@@ -1594,7 +1592,7 @@ export class ComponentBrowserProvider {
           ${component.documentationUrl ?
             `<div><strong>Project URL:</strong> <a href="${component.documentationUrl}" target="_blank" id="componentDocUrl">${component.documentationUrl}</a></div>` : ''}
           ${component.url ?
-            `<div><strong>Component URL:</strong> <a href="${component.url}" target="_blank" id="componentUrl">${component.url}</a></div>` : ''}
+            `<div><strong>Component URL:</strong> <code id="componentUrl">${component.url}</code></div>` : ''}
           ${templateFileUrl ?
             `<div><strong>Template File:</strong> <a href="${templateFileUrl}" target="_blank" id="templateFileUrl">${templateFileUrl}</a></div>` : ''}
         </div>
@@ -1745,14 +1743,6 @@ export class ComponentBrowserProvider {
             toggleButton.textContent = isHidden ? 'Hide' : 'Show';
           }
 
-          function buildTemplateFileUrlFromComponent(component) {
-            if (!component || !component.gitlabInstance || !component.sourcePath || !component.name) {
-              return '';
-            }
-            const ref = component.version || 'main';
-            return 'https://' + component.gitlabInstance + '/' + component.sourcePath + '/-/blob/' + ref + '/templates/' + component.name + '.yml';
-          }
-
           function updateComponentDetails(component) {
             console.log('Updating component details:', component);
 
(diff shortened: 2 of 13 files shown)


eFAILution (commented):
Thanks Simon, this fixes a real bug and the root-cause fix is correct: the old hardcoded `/-/blob/<ref>/templates/<name>.yml` 404s for directory-layout repos, and routing single-file vs directory through the recorded `templatePath` is the right approach. The DocumentLinkProvider, the `$CI_SERVER_FQDN` expansion, and the no-link-when-unconfigured behavior all look sound.

The main gap is `templatePath` propagation. It's populated on the catalog-discovery path, but a couple of paths leave it undefined:
- The dynamic/per-URL fetch cache writes don't carry it (`componentDetector.ts:318-327` and `:632-655`).
- The catalog branch of `fetchComponentMetadata` returns without it (`componentFetcher.ts:189-206`).

So links work for catalog-discovered components but go missing for ones that enter the cache via dynamic fetch. It fails gracefully (no link rather than a 404), but it undercuts the "every entry gets a link" intent, and there's no test on the propagation chain. Adding `templatePath` to those payloads would close it.

On the cache bump: dropping the whole cache on 1.0.0 -> 1.1.0 is safe in both directions, but it's heavier than needed for an optional cosmetic field that both consumers already treat as optional. A lazy backfill (or just letting normal TTL expiry repopulate it) would avoid forcing every user to re-fetch from GitLab on upgrade. Your call, not blocking.

Minor: `documentLinkProvider.ts:314-319` adds another copy of `isGitLabCIFile` that omits the `gitlab-ci` languageId clause the other providers have. Worth converging these eventually (related to the matcher work in #125).

Process note: this is a fix, so if it's meant as an immediate patch to the current stable line, targeting `main` is deliberate, keep it. Otherwise it should target `beta` and ship with the rest of the next release.

PR #167, opened 2026-06-10
Chore: eslint fix phase 7

X-Guardian:
## Summary
- Phase 7 — the final piece of the ESLint `@typescript-eslint/no-explicit-any` cleanup. Tackles the Component Browser family (`componentBrowserProvider.ts`, `componentBrowserTransform.ts`, `componentBrowserGenerate.ts`) with typed hierarchy shapes (`SourceGroup` → `ProjectGroup` → `ComponentGroup` → `ComponentVersion`), narrow function-local component shapes for the YAML generator, and runtime type-guards for the webview detach round-trip.
- **41 → 0 `no-explicit-any` warnings**, ending the cleanup arc (started at 218 across the 7 phases).
- `npm run lint` now runs with `--max-warnings 0`, scans the whole repo (not just `src/`), and the ESLint config got a structural cleanup along the way: scoped presets, narrower `ignores` list, `globals.node`, dropped redundant config, picked up `reportUnusedDisableDirectives: 'warn'`.
- Two latent runtime bugs surfaced and fixed (see below).
- Link related issue(s): n/a

## Change Type
- [ ] feat
- [ ] fix
- [ ] refactor
- [ ] docs
- [ ] test
- [x] chore

### User-facing impact
- None expected from the typing edits themselves. Two latent bugs were fixed in passing: error dialogs in the Component Browser now show the real error message instead of `[object Object]`; the context-source dedup check no longer always triggers a redundant re-fetch.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (no behavioural change to either)

### Affected areas
- [x] Component Browser (typing + two latent bug fixes — see below)
- [ ] Hover provider
- [ ] Completion provider
- [ ] Validation provider
- [ ] Docs only

## What changed by bucket

| Bucket | Files | Warnings cleared | Approach |
|---|---|---|---|
| Hierarchy types | [componentBrowserTypes.ts](src/providers/componentBrowserTypes.ts) (new) | (foundation) | Two type families: **builder shapes** (`SourceGroupBuilder` / `ProjectGroupBuilder` / `ComponentGroupBuilder` — carry `Map<string, …>` during transform) and **tree shapes** (`SourceGroup` / `ProjectGroup` / `ComponentGroup` / `ComponentVersion` — the flat-array form returned to the webview). Per-property JSDoc on every field. |
| Transform | [componentBrowserTransform.ts](src/providers/componentBrowserTransform.ts) | 9 | `transformCachedComponentsToGroups(cachedComponents: Array<Partial<CachedComponent>>): SourceGroup[]`. Maps typed at every step; the flatten pass infers cleanly. `Partial<CachedComponent>` is honest — the runtime's `if (!comp.source \|\| !comp.sourcePath \|\| !comp.name) continue;` is the actual contract. |
| Generator | [componentBrowserGenerate.ts](src/providers/componentBrowserGenerate.ts) | 6 | Local `GeneratableComponent` / `GeneratableParameter` / `ExistingComponent` types declare exactly what the routine reads (no `description` — never used; `default: unknown` — `JSON.stringify` accepts arbitrary shapes). Honest about being looser than `ComponentParameter`. |
| Provider | [componentBrowserProvider.ts](src/providers/componentBrowserProvider.ts) | 26 | `Component` / `DetachableComponent` on the public methods; typed `SourceGroup[]` on the HTML builders; two runtime type-guards (`isCachedComponentShape`, `isExistingComponentShape`) replace `as` casts; `getErrorHtml(error: unknown)` narrows via `instanceof Error`; dead `fetchSpecificVersion` method deleted. |
| Lint gate + config cleanup | [eslint.config.js](eslint.config.js), [package.json](package.json) | (rule enforcement) | `"lint": "eslint --max-warnings 0"` (no path arg — scans the whole repo, including tests, against the ignores list). Multiple structural fixes in `eslint.config.js`: see "ESLint config cleanup" below. |
| Test fixture corrections | [tests/unit/*](tests/unit/) | 4 errors + 1 warning surfaced once `tests/**` came out of the ignore list | One `default?: any` → `unknown`, three `\$` and `\)` escape removals (escapes that were no-ops in their respective context). One missing `description` field added to a `ComponentParameter` fixture. |
| Release script cleanup | [scripts/*.js](scripts/) | 4 errors + 10 warnings surfaced once `scripts/**` came out of the ignore list | Removed two obsolete `/* eslint-env node */` comments (flat-config replaces env comments with `languageOptions.globals`); dropped 8 stale `// eslint-disable[-next-line] no-process-exit` directives (the rule isn't enabled now); replaced two unused `catch (error)` with `catch {}`; removed two `let changelogContent = ''` initialisers that were unconditionally overwritten in both branches of the next `if`/`else`. |

Total: **41 `no-explicit-any` warnings removed.** Repo-wide `eslint .` now passes with zero errors and zero warnings — the first time in the cleanup arc.

### Builder shapes vs tree shapes
The transform builds a tree via `Map<string, …>` for dedup-by-key, then flattens to arrays for the webview. Separating these into two family of types (`*Builder` vs the unsuffixed forms) keeps the internal data structure honest about what's a Map and what's an Array, without forcing the webview-facing type to carry Maps it never sees.

### `GeneratableComponent` instead of `Component`
The browser feeds two flavours into `generateComponentText`: a cached catalog component and a freshly-detected one from the editor. Neither matches the canonical `Component` shape exactly — and unit tests pass partial fixtures (e.g. `default: { key: 'value' }` — outside `ParameterDefault`'s union). The local type captures exactly the fields read (`name`, `version`, `sourcePath`, `gitlabInstance`, `originalUrl`, `parameters[]`); `default: unknown` honestly admits the generator works on whatever `JSON.stringify` can encode.

### Runtime type-guards over `as` casts
`isCachedComponentShape(component)` checks the five required cache fields before passing the value into `cacheManager.fetchComponentVersions`; `isExistingComponentShape(value)` narrows the `unknown` return of `parseExistingComponentText` to `{ inputs?: Record<string, unknown> }`. Pattern matches Phase 4's `isYamlNode` and Phase 6's `isIncludeEntry`.

### `DetachableComponent` mirrors `extension.ts`
The detach-hover webview carries a `_hoverContext` field added at runtime by `hoverContentBuilder`. Both the extension entry and the browser receive these shapes — same `Component & { _hoverContext?: HoverContext }` alias declared in both, kept local rather than centralised because the field is a runtime-only extension to the canonical model.

## ESLint config cleanup

`eslint.config.js` had accumulated structural noise over the cleanup arc. This PR fixes:

- **Scoped presets to where they belong.** `js.configs.recommended` and `tseslint.configs.recommended` were both spread at the root with no `files:` filter — so the TS preset was applying its rules (including `no-require-imports`) to CommonJS configs like `eslint.config.js` itself, and the JS preset was layering on top of the TS preset for `.ts` files. Now: TS preset wrapped in `tseslint.config({ files: ['**/*.ts'], extends: [...] })`; JS preset spread into the `**/*.js`/`**/*.cjs` block's `rules`.
- **Narrowed the `ignores` list** from 17 → 6 entries. Dropped redundant globs (`**/*.json`, `**/*.yaml`, `**/*.yml`, `**/*.md`, `**/*.sh` — ESLint doesn't process those extensions anyway), a non-existent directory (`dist/**`), a file already covered by the `**/*.js` block (`esbuild.js`), an over-ignore that hid 4 errors + 1 warning in the test suite (`tests/**`), and `scripts/**` (which uncovered 14 issues fixed in this PR).
- **Replaced the hand-rolled Node globals map** (`console`, `URL`, `module`, `require`, …) with `globals.node` from the `globals` package (new devDep, ~5KB, maintained by the ESLint org). Future scripts using `setTimeout`, `URLSearchParams`, etc. no longer need new entries.
- **Added `linterOptions.reportUnusedDisableDirectives: 'warn'`** so stale `// eslint-disable` directives surface as warnings — caught the 8 dead `no-process-exit` disables in `scripts/`.
- **Added `@type {import('eslint').Linter.Config[]}` JSDoc** so editors check the config against ESLint's published schema.
- **Dropped redundant `languageOptions.ecmaVersion: 'latest'` and `sourceType: 'module'`** — both are ESLint v9 defaults.
- **Dropped two dead rule overrides** in the TS block: `'no-unused-vars': 'off'` (the TS preset already disables it) and `curly: 'off'` (no preset enables it). Also dropped `'@typescript-eslint/no-explicit-any': 'warn'` — the cleanup is complete and the preset's `'error'` default is now the right severity.
- **Important constraint flagged in a comment**: ESLint flat-config treats `{ ignores: […] }` as a global ignore *only* when `ignores` is the sole key — adding any other key downgrades it to a regular config block. Adding `linterOptions` alongside `ignores` accidentally broke this and made the linter walk `.vscode-test/**`. Now they're split into two separate top-level blocks with a comment explaining why.
- **Deleted the legacy `.eslintrc.js`**. It was the old eslintrc-format config superseded by the flat `eslint.config.js`.

### `getErrorHtml` rendered `[object Object]`
The catch handler at [componentBrowserProvider.ts:340-342](src/providers/componentBrowserProvider.ts#L340-L342) passes the caught error into `getErrorHtml(error: any)`, which interpolated `${error}` into the rendered HTML. For any `Error` instance — which is what `componentService.fetchCatalogData` and friends throw — the template produced `[object Object]` in the user-visible error dialog. Now uses `error instanceof Error ? error.message : String(error)` (the same pattern as Phase 2's `extractMessage`).

### `contextSourceExists` was always `false`
[componentBrowserProvider.ts:218-220](src/providers/componentBrowserProvider.ts#L218-L220) checked `group.gitlabInstance === contextInstance && group.sourcePath === contextPath` — but `SourceGroup` doesn't carry those fields (they live on individual projects under the source). The check always returned `false` at runtime, which meant the context source was *always* re-fetched and prepended even when it was already in the cache. Fix: check via `group.projects.some(p => …)`. The redundant fetch on cache-hit now goes away.

### `selectDefaultVersion(component.version)` always passed `undefined`
[componentBrowserTransform.ts](src/providers/componentBrowserTransform.ts) read `component.version` on the *builder* object, which only has `defaultVersion` (no `version` field). The fallback `|| 'latest'` always engaged. Now passes `component.defaultVersion`, so the default-version selection uses the right initial value.

### Project-level `gitlabInstance` was inconsistently defaulted
The transform defaulted `gitlabInstance` to `'gitlab.com'` at the component level but not the project level — projects could receive `undefined`. Now both levels default consistently.

### Local checks
- [x] `npm run check-types` — clean (both `tsconfig.json` and `tsconfig.tests.json`).
- [x] `npm run lint` — **0 warnings, 0 errors repo-wide** (gate is `--max-warnings 0`, scans the whole repo).
- [x] `npm test` — 206/206 unit assertions passing.
- [x] Manual verification in VS Code Extension Host — `env -u ELECTRON_RUN_AS_NODE npm run test:extension-host` 10/10 passing.

### Manual test notes
- The component browser's tree rendering exercises the new `SourceGroup → ProjectGroup → ComponentGroup → ComponentVersion` chain end-to-end on every "show browser" call.
- The detach-hover round-trip (hover → "Open in Detailed View" → insert/edit) exercises `DetachableComponent` and both runtime type-guards.
- The "Edit existing component" path exercises `parseExistingComponent → isExistingComponentShape → generateComponentText`.
- The context-source dedup fix can be observed by hovering a component referenced in the file, opening the browser, and confirming there's no extra network call to re-fetch the catalog for that source (it was previously refetched every time because the `contextSourceExists` check always returned `false`).

### Test fixture updates
- `tests/unit/componentBrowserTransform.test.ts` — added a missing `description` field on a `ComponentParameter`; `Array<{ comp: any; … }>` → `Array<{ comp: unknown; … }>` (the `onSkip` callback already types `component` as `unknown`).
- `tests/unit/GitLabSpecParser.test.ts` — two unnecessary `\$` escapes in template literals removed (`$` is only special before `{` — `\$[[…]]` and `$[[…]]` are the same string).
- `tests/unit/hoverContentBuilder.test.ts` — two unnecessary `\)` escapes inside regex character classes removed (`[^\)]` is the same as `[^)]`).

### Release-script cleanups
- `scripts/manual-release.js` and `scripts/version-bump.js` — removed obsolete `/* eslint-env node */` comments (flat config doesn't read them), 4 dead `// eslint-disable-line no-process-exit` directives, 2 unused `catch (error)` parameters (now `catch {}`), and 2 `let changelogContent = ''` initialisers that were unconditionally overwritten in both arms of the next `if`/`else`.
- `scripts/release.js` — removed 4 dead `// eslint-disable-next-line no-process-exit` directives.

## Breaking Changes
- [x] No breaking changes

The hierarchy types (`SourceGroup`, `ProjectGroup`, etc.) are internal to the extension — they describe what the webview consumes, not any persisted format or public API.

- N/A — no UI changes. Two latent rendering fixes (real error messages in the error dialog, no redundant network fetches) are described under "Latent runtime bugs fixed".

## Risk and Rollback
- **Main risks:**
  - The `contextSourceExists` fix changes a "re-fetch every time" behaviour into "fetch only on cache miss". Net safer — fewer network calls — but a user who'd been implicitly relying on the redundant fetch to refresh stale catalog data would no longer get it. The branch-freshness logic from [#153] already covers that case for the live entries it handles; for catalog entries the normal refresh paths still apply.
  - `--max-warnings 0` on the lint script means any new lint warning (in any rule) blocks both local `npm run lint` and CI. The cleanup pre-cleared all warnings; future contributors will see the gate as a guardrail rather than a surprise.
- **Rollback strategy:** Single PR, revert. One new devDep (`globals`) but no settings changes, no data migration, no CI secrets touched, no new runtime dependencies.

## Release Notes Draft
- chore: complete the ESLint `no-explicit-any` cleanup arc — Component Browser hierarchy and YAML generator are now strictly typed; `npm run lint` runs repo-wide with `--max-warnings 0`. ESLint config simplified along the way (scoped presets, `globals` package, dropped redundant settings). Drive-by fixes: error dialogs in the Component Browser now show real error messages; the context-source dedup no longer always triggers a redundant catalog re-fetch.

## Checklist
- [x] Added/updated docs for behavior or settings changes (none required)
- [x] Added/updated tests for new behavior (test fixtures corrected; behaviour unchanged)


package.json
```diff
@@ -275,7 +275,7 @@
     "watch:tsc": "tsc --noEmit --watch --project tsconfig.json",
     "package": "npm run check-types && node esbuild.js --production",
     "check-types": "tsc --noEmit && tsc --noEmit --project tsconfig.tests.json",
-    "lint": "eslint src",
+    "lint": "eslint --max-warnings 0",
```
eFAILution, line 278: `--max-warnings 0` with no path arg is now a repo-wide gate: any new warning in any rule fails both local `npm run lint` and CI for every subsequent PR. Intentional and a good guardrail given the cleanup pre-cleared everything — just flagging that #169 (which adds `tests/**` code, now lintable since this PR un-ignores that dir) must be warning-clean against this config once both land.

eFAILution (commented):
## Review summary

Approve from my side — large but low-risk. I verified the four latent bug fixes against the new types and they're all real:

- `getErrorHtml` `[object Object]` → `instanceof Error` narrowing. ✓
- `contextSourceExists` checked fields `SourceGroup` doesn't carry; now `projects.some(…)`. ✓
- `selectDefaultVersion` read `version` off the builder (which only has `defaultVersion`); fixed. ✓
- `let changelogContent;` is safe — both `if`/`else` arms assign before any read, and the no-commits path returns early. ✓

No `: any` remain in the browser family, and the flat-config `ignores`/`linterOptions` split is correct (the global ignore stays a sole-key block). One note inline on the lint gate.

**Sequencing:** whichever of #167/#169 merges second inherits this repo-wide `--max-warnings 0` gate, and this PR pulls `tests/**` out of the ignore list — worth a `npm run lint` on the merged result rather than trusting each PR's in-isolation green check.

Issue #170, opened 2026-06-11
feat: scope component versions per-component in monorepos


X-Guardian:
### Problem statement

When a single GitLab project holds **many** CI components, releases are usually distinguished by tags that embed the component name — e.g. `deploy-app-1.1.0`, `deploy-app-2`, `build-image-4.0.0`. One project therefore carries the tags of *every* component mixed together.

The extension is currently monorepo-unaware. Version discovery fetches the project's entire tag list and assigns it to **every** component, so:

- Each component's version dropdown lists hundreds of unrelated tags (every other component's releases), not just its own.
- The "N versions" badge is meaningless (it counts the whole repo's tags).
- Semantic-version sorting and default-version selection mix versions across components.
- The same happens in the **details** panel and the **editor-hover** details panel — both show the full unscoped tag list, unstripped.

For a real monorepo with dozens of components, this makes the version picker effectively unusable.

### Proposed solution

Let a component source declare how its tags embed the component name, then scope each component's version list to its own tags.

Add a per-source **tag-version template** — a string with two tokens:

- `{name}` — the component (= `templates/` directory) name.
- `{version}` — the version surfaced in the dropdown; captured for display. Matches anything starting with a digit.

The template compiles to a regex (literals escaped, `{name}` → the component name, `{version}` → a capture group). Each component's dropdown is scoped to the tags that match, and labels are shown stripped of the prefix (`1.1.0`, not `deploy-app-1.1.0`). The **full tag** is retained as the value inserted into the file, so the GitLab include still resolves.

Behaviour is **opt-in per source**: setting `tagPattern` enables monorepo scoping; leaving it unset means an ordinary single-component repo, unchanged.

Declare the convention on the source (`settings.json`):

```jsonc
"gitlabComponentHelper.componentSources": [
  {
    "name": "Shared CI Monorepo",
    "path": "my-group/shared-ci",
    "gitlabInstance": "gitlab.com",
    "tagPattern": "{name}-{version}"
  }
]
```

Other conventions are expressible by changing the template:

| Tag style | Template |
|---|---|
| `deploy-app-1.1.0` | `{name}-{version}` |
| `apps/web/v2.0.0` | `apps/{name}/v{version}` |
| `web_1.0.0` | `{name}_{version}` |

### Example .gitlab-ci.yml usage

```yaml
With the source configured as above, picking `deploy-app` 2.1.0 from the (now scoped) dropdown inserts the full tag as the include ref:


include:
[...]

PR #171, opened 2026-06-11
feat: monorepo tag scoping

X-Guardian:
## Summary
- Makes version discovery **monorepo-aware**. In a tag-per-component monorepo, one project carries every component's tags (e.g. `deploy-app-1.1.0`, `build-image-4.0.0`). Previously every component's dropdown listed *all* repo tags, unscoped and unstripped, with a meaningless version count and broken default selection.
- **New per-source tag-version template.** A source can declare `tagPattern` (e.g. `{name}-{version}`, `apps/{name}/v{version}`, `{name}_{version}`) using two tokens: `{name}` (the component/directory name) and `{version}` (the version shown; matches anything starting with a digit). Its presence enables monorepo scoping; leaving it unset means an ordinary single-component repo, unchanged.
- **New pure matcher** (`src/services/component/tagScoping.ts`): `compileTagTemplate(template, name)` compiles a template to an anchored regex (literals escaped, `{name}` → the escaped component name, `{version}` → a `(\d.*)` capture). `scopeTagsToComponent` and `stripTagPrefix` delegate to it. The digit anchor stops a component whose name is a prefix of another's (`build-image` vs `build-image-extra`) from capturing the sibling's tags.
- **Wired through discovery and UI**: the template flows from the source config → `CachedComponent` → version discovery (`versionCache.ts`), default-version scoring (`componentBrowserTransform.ts`), and all three version dropdowns — Component Browser, browser **Details** panel, and editor-**hover** details panel (`componentBrowserProvider.ts`, `extension.ts`). Option **values** stay the full tag (the inserted ref); only **labels** are stripped.
- **Default-branch fix**: the dropdown previously hard-coded `main` *and* `master`. It now lists only the repo's real `default_branch` (new `fetchProjectDefaultBranch`), so phantom branches no longer appear.
- Link related issue(s): Fixes #170

## Change Type
- [x] feat
- [x] fix
- [ ] refactor
- [ ] docs
- [x] test
- [ ] chore

### User-facing impact
- Sources flagged as monorepos show each component **only its own versions**, prefix-stripped (`2.1.0`, not `deploy-app-2.1.0`), with an accurate version count and a sensible default (highest semver of that component).
- The inserted include ref is unchanged in form — the full tag, e.g. `…/deploy-app@deploy-app-2.1.0`.
- Scoping is consistent across the Component Browser, the browser details panel, and the editor-hover details panel.
- The version dropdown no longer lists a `master` branch that doesn't exist; it shows the repo's real default branch.
- No change for ordinary single-component sources (no `tagPattern`).

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (template/scoping logic is instance-independent; default branch comes from the standard project-info API)

### Affected areas
- [x] Component Browser
- [x] Hover provider
- [ ] Completion provider
- [ ] Validation provider
- [x] GitLab API calls/auth/token storage
- [ ] Docs only

### Local checks
- [x] `npm run compile`

### Manual test notes
Against a tag-per-component monorepo (dozens of components) with:
```jsonc
"gitlabComponentHelper.componentSources": [
  {
    "name": "Shared CI",
    "path": "my-group/shared-ci",
    "gitlabInstance": "gitlab.com",
    "tagPattern": "{name}-{version}"
  }
]
```
Observed after **Reset Cache** + reload:
- `deploy-app` shows only its `1.x`/`2.x` tags (labelled `2.1.0`, `2.0.0`, …); `build-image` shows only its `4.x` tags. Counts match.
- Inserting `2.1.0` writes `…/deploy-app@deploy-app-2.1.0`.
- The **Details** panel and the **editor-hover** details panel (triggered from an `include:` in a `.gitlab-ci.yml`) both show the same scoped, stripped list.
- Only the real default branch appears (no `master`).
- Custom templates exercised in unit tests: `apps/{name}/v{version}`, `{name}_{version}`.
- A source without `tagPattern` still lists all tags exactly as before.

Unit tests in `tests/unit/tagScoping.test.ts` cover the default and custom templates, regex-metacharacter escaping, the digit-anchor sibling-collision guard, malformed templates (missing `{version}` → no match), and template-aware default selection. Full suite: 221 passing.

## Breaking Changes
- [x] No breaking changes

New settings are additive and opt-in. The persisted version-cache shape gains a `defaultBranches` map, but `deserializeCache` still reads the previous tags-only array form, so no cache-version bump is needed and no user action is required.

## Risk and Rollback
- Main risks:
  - A custom template that's too greedy could over-match a sibling component's tags. Mitigated by the digit-anchored `{version}` and documented in the README.
  - One extra lightweight API call per project to read `default_branch` (cached per project alongside tags).
- Rollback strategy: revert the PR. The settings are opt-in, so reverting only removes the scoping; no data migration. A reverted build's `deserializeCache` can't read this PR's object-shaped snapshot, but that failure is caught in `loadCacheFromDisk`'s try/catch — the cache is treated as empty and refetched on next load, so the revert is safe.

## Release Notes Draft
- Monorepo support: scope each component's version dropdown to its own tags via a per-source `tagPattern`; the version picker now also shows only the repo's real default branch.

## Checklist

### DIFF
--- a/src/providers/componentBrowserProvider.ts
+++ b/src/providers/componentBrowserProvider.ts
@@ -1185,8 +1185,17 @@ export class ComponentBrowserProvider {
                 // full tag as the option value (the inserted ref).
                 const labels = message.versionLabels || {};
                 const label = function(v) { return labels[v] || v; };
-                const opts = versions.map(function(v) { return '<option value="' + v + '" ' + (v === defaultVersion ? 'selected' : '') + '>' + label(v) + '</option>'; }).join('');
-                actionsDiv.innerHTML = '<select class="version-dropdown" onchange="updateComponentVersion(&#39;' + componentName + '&#39;, this.value, &#39;' + projectId + '&#39;)">' + opts + '</select><button onclick="viewDetailsById(&#39;' + componentName + '&#39;, &#39;' + defaultVersion + '&#39;, &#39;' + projectId + '&#39;)">Details</button><button onclick="insertComponentById(&#39;' + componentName + '&#39;, &#39;' + defaultVersion + '&#39;, &#39;' + projectId + '&#39;)">Insert</button>';
+                // Build the dropdown shell + buttons as markup, then append options via the DOM so the untrusted
+                // version strings (tag names can contain <, >, &) are never interpolated into HTML.
+                actionsDiv.innerHTML = '<select class="version-dropdown" onchange="updateComponentVersion(&#39;' + componentName + '&#39;, this.value, &#39;' + projectId + '&#39;)"></select><button onclick="viewDetailsById(&#39;' + componentName + '&#39;, &#39;' + defaultVersion + '&#39;, &#39;' + projectId + '&#39;)">Details</button><button onclick="insertComponentById(&#39;' + componentName + '&#39;, &#39;' + defaultVersion + '&#39;, &#39;' + projectId + '&#39;)">Insert</button>';
+                const select = actionsDiv.querySelector('.version-dropdown');
+                versions.forEach(function(v) {
+                  const option = document.createElement('option');
+                  option.value = v;
+                  option.textContent = label(v);
+                  if (v === defaultVersion) { option.selected = true; }
+                  select.appendChild(option);
+                });
                 const descElement = document.getElementById('desc-' + componentName + '-' + projectId);
                 if (descElement && !document.getElementById('version-info-' + componentName + '-' + projectId)) {
                   const versionInfo = document.createElement('div');
@@ -1744,7 +1753,7 @@ export class ComponentBrowserProvider {
                   : null;
                 return availableVersions.map((version: string) => {
                   const label = matcher?.extractVersion(version) ?? version;
-                  return `<option value="${version}" ${version === component.version ? 'selected' : ''}>${label}</option>`;
+                  return `<option value="${this.escapeHtml(version)}" ${version === component.version ? 'selected' : ''}>${this.escapeHtml(label)}</option>`;
                 }).join('');
               })()}
             </select>
--- a/src/providers/componentBrowserProvider.ts
+++ b/src/providers/componentBrowserProvider.ts
@@ -12,6 +12,7 @@ import { templateFileUrlForResolved } from '../utils/templateFileUrl';
 import { generateComponentText } from './componentBrowserGenerate';
 import { findComponentLineRange, parseExistingComponentText } from './componentBrowserEdit';
 import { transformCachedComponentsToGroups } from './componentBrowserTransform';
+import { compileTagTemplate, stripTagPrefix } from '../services/component/tagScoping';
 
 /**
  * Component shape carried through the detach-hover webview's "Open in Detailed View" round trip.
@@ -492,8 +493,12 @@ export class ComponentBrowserProvider {
       }
     );
 
+    // The component arrives from the webview carrying only the selected version and none of the source settings;
+    // recover the full version list and monorepo settings from the cache so the dropdown is complete and labelled.
+    const enriched = await this.lookupComponentDetails(component);
+
     // Show component details
-    detailsPanel.webview.html = this.getComponentDetailsHtml(component);
+    detailsPanel.webview.html = this.getComponentDetailsHtml({ ...component, ...enriched });
 
     // Handle messages from the details panel
     detailsPanel.webview.onDidReceiveMessage(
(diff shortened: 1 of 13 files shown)


src/services/cache/versionCache.ts
```diff
@@ -107,43 +142,44 @@ export class VersionCache {
...
 
   /**
    * Get serializable cache data for persistence
    *
-   * @returns Array of [key, versions] tuples
+   * @returns Array of [key, tags] tuples
    */
   serializeCache(): Array<[string, string[]]> {
-    return Array.from(this.projectVersionsCache.entries());
+    return Array.from(this.projectTagsCache.entries());
```
eFAILution: **The default branch disappears from the version dropdown after a cache restore — a regression.**

`serializeCache` persists only the raw tags; `projectDefaultBranchCache` is never serialized. So on a restored session:

1. `loadCacheFromDisk` → `deserializeCache` hydrates `projectTagsCache`, leaving `projectDefaultBranchCache` empty.
2. The first `fetchComponentVersions` is a `projectTags` **cache hit** → the hit branch never fetches the default branch → the read at line 64 (`projectDefaultBranchCache.get(cacheKey)`) is `undefined` → `branches = []` → no branch in the dropdown.

Since `globalState` persists across sessions, this is the steady-state path, not an edge case. The branch shows up only on the first post-fetch session, then vanishes on every restart. The previous code baked the sorted `main`/`master`/tags list straight into the serialized blob, so branches survived restore — splitting tags from the branch lost that.

Fix direction: persist the default-branch map alongside the tags (extend the serialized shape + `deserializeCache`), or resolve the default branch lazily when a tags-hit finds no cached branch entry.

src/services/component/versionManager.ts
```diff
@@ -21,11 +22,17 @@ export class VersionManager {
    * Fetch all tags/versions for a GitLab project with optimizations
    * @param gitlabInstance The GitLab instance hostname
    * @param projectPath The project path
+   * @param scopeToComponent When set, the project is treated as a tag-per-component monorepo and tags are scoped to
+   *                         this component using `tagPattern` (full tags retained as the version strings).
+   * @param tagPattern The tag-version template for scoping (e.g. `{name}-{version}`); defaults to the house
+   *                           convention when omitted.
    * @returns Array of version strings (tags and important branches)
    */
   public async fetchProjectVersions(
```
eFAILution, line 26: **Low priority / cleanup:** the new `scopeToComponent` / `tagPattern` params on `fetchProjectVersions` (here and on the `componentService` wrapper) are dead — nothing in `src/` or `tests/` calls `fetchProjectVersions`. The live path is `versionCache.fetchComponentVersions` → `fetchProjectTags` + inline `scopeTagsToComponent`. Worth dropping the unused params (and their docstrings) to avoid implying a second scoping entry point.

src/providers/componentBrowserProvider.ts
```diff
@@ -1648,9 +1736,17 @@ export class ComponentBrowserProvider {
...
-                `<option value="${version}" ${version === component.version ? 'selected' : ''}>${version}</option>`
-              ).join('')}
+              ${(() => {
+                // For monorepo tags show the template's {version} capture as the label, keeping the full tag as the
+                // option value (the ref used to fetch and insert the version).
+                const matcher = component.tagPattern
+                  ? compileTagTemplate(component.tagPattern, component.name)
+                  : null;
+                return availableVersions.map((version: string) => {
+                  const label = matcher?.extractVersion(version) ?? version;
```
eFAILution, line 1755: **Low priority / hardening nit:** this details-panel dropdown interpolates `value="${version}"` and `>${label}<` unescaped, as does the browser-tree async builder (~line 1180), while the new `renderVersionOptions` (line 668) correctly uses `escapeHtml`. Git tag names can legally contain `<`/`>`/`&`, so a tag in a configured source is a (low-likelihood, webview-CSP-mitigated) injection surface. The `version` interpolation predates this PR, but since you're adding `label` to these paths and already escape it one method over, worth making all three consistent.

eFAILution (commented):
## Review

The core is strong — `tagScoping.ts` is a clean, pure, well-tested matcher (anchored regex, escaped literals, digit-anchored `{version}` to stop sibling-prefix collisions), and the scoping/default-selection logic is correct. One functional regression worth fixing before merge, plus two low-priority notes; all inline.

(Re the cache-version bump the PR body mentions — not a concern: #169 already moved beta to `1.3.0`, so the eventual beta→main release will mismatch users' installed cache version and refresh regardless. No per-PR bump needed.)

eFAILution (approved):
Approving — all three findings resolved and verified against the `Resolve review comments` commit:

- **Default branch on restore (regression):** fixed via `VersionCacheSnapshot` carrying both `tags` and `defaultBranches`; `serializeCache`/`deserializeCache` round-trip both and the legacy array form is still accepted on read. Clean, backward-compatible.
- **Dead `fetchProjectVersions` scoping params:** removed from both layers plus the unused import.
- **Dropdown HTML escaping:** consistent across all three paths now (`escapeHtml` on value+label; async builder moved to `createElement`).

CI green (unit + extension-host + CodeQL + security scans). Nice work.

Issue #184, opened 2026-06-18
bug: input completions offered outside the input-name slot


X-Guardian:
### Bug summary

When editing an include's `inputs:` block, the extension offers input-name completions in positions that are not a valid parameter-name slot:

1. **Inside a multi-line array value** — completions appear on an array item line (`- value`) under an input.
2. **At the wrong indentation** — completions appear one column shallower than the existing input keys, and several columns deeper than them.
3. **When the cursor is moved left** — after typing the correct indentation for a new input, moving the cursor left along the leading whitespace still shows completions.

### Steps to reproduce

1. Open a `.gitlab-ci.yml` with an include that has an `inputs:` block (see the snippet below).
2. Trigger completion in each of these positions:
   - On an array item line (`- tag-b`) under a multi-line array input.
   - One space to the left of the existing input keys, and several spaces to the right of them.
   - Type the correct indentation for a new input, then move the cursor left along the whitespace.
3. Observe that input-name suggestions are offered in all of these positions.

### Expected behavior

Input-name completions are offered only when the cursor sits in a genuine parameter-name slot: aligned with the existing input keys' column (or, for an empty `inputs:` block, deeper than `inputs:`), not on an array-item line, and with the cursor at or beyond the slot indent — never left of it.

### Actual behavior

Completions are offered on array-item lines, at misaligned indentation (shallower or deeper than the existing keys), and when the cursor has been moved left of the slot indent.

### Minimal .gitlab-ci.yml snippet

```yaml
include:
  - component: $CI_SERVER_FQDN/group/project/my-component@1.0.0
    inputs:
      tags:
        - tag-a
        - tag-b          # (1) completion offered on this array item line
      environment: dev   # input keys are at 6 spaces
                         # (2) completion offered at 5 spaces (left) and 8+ spaces (right of the keys)
                         # (3) type 6 spaces for a new input, move cursor left → still offered
```

### Extension logs

```text
N/A
```

### Affected area

Completion provider

### Extension version

0.13.2

### VS Code version

1.124.2

### Node.js version

22

### Operating system

macOS

### Pre-submit checks

- [x] I searched existing issues and did not find a duplicate.
- [x] I can reproduce this on the latest version from main/beta.
- [x] I included a minimal YAML snippet and relevant logs.

PR #185, opened 2026-06-18
fix: input completions offered outside the input-name slot

X-Guardian:
## Summary
- Fixes input-name completion being offered outside a genuine parameter-name slot. Slot detection in `findCompletionInputContextAtLine` decided "is this a name slot?" from the line's leading-whitespace count alone (just "deeper than `inputs:`"), so it fired on array-item lines, at misaligned indentation (shallower or deeper than the existing keys), and — because it never read the cursor column — when the cursor was moved left along the indentation of an otherwise-correct line.
- Anchors the slot to the column the existing input keys line up at, excludes array-item lines, and gates on the cursor column so the cursor must sit at or beyond the slot indent.
- Link related issue(s): Fixes #184 

## Change Type
- [ ] feat
- [x] fix
- [ ] refactor
- [ ] docs
- [ ] test
- [ ] chore

### User-facing impact
- Input-name suggestions now appear only in a real parameter-name slot: aligned with the existing input keys (or, for an empty `inputs:` block, deeper than `inputs:`), never on an array value's `- item` line, and only when the cursor actually sits in the slot rather than having been moved left along its leading whitespace.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (local slot detection only; no GitLab interaction)

### Affected areas
- [ ] Component Browser
- [ ] Hover provider
- [x] Completion provider
- [ ] Validation provider
- [ ] Docs only

## What changed by bucket

| Bucket | Files | Approach |
|---|---|---|
| Slot indentation | [completionInputContext.ts](../src/providers/completionInputContext.ts) | `findInputsSection` now also reports `childIndent` — the column of the first real input key under `inputs:` (skipping array items and the `inputs:` header). The slot check requires the cursor line to sit at exactly that column, falling back to the old "deeper than `inputs:`" rule only for an empty `inputs:` block with no keys to align to. Array-item lines (`- value`, bare `-`) are excluded, since they belong to a parameter's value, not a name slot. |
| Cursor column | [completionInputContext.ts](../src/providers/completionInputContext.ts), [completionProvider.ts](../src/providers/completionProvider.ts) | `findCompletionInputContextAtLine` takes an optional `column`; when supplied, the cursor must sit at or beyond the slot indent (`column >= currentIndent`), never left of it. The caller now passes `position.character` through. Omitting `column` preserves the cursor-agnostic behaviour the unit tests use. |
| Regression tests | [completionInputContext.test.ts](../tests/unit/completionInputContext.test.ts) | Added cases: array-item line → null; slot one column shallower → null; slot deeper/nested → null; slot aligned with the keys → detected; cursor moved left of the slot indent → null; cursor at the slot indent → detected. |

### Local checks
- [x] `npm run compile` (`tsc --noEmit` clean)
- [x] `npm test` (261 passing)
- [x] `npm run lint` clean

### Manual test notes
- Reproduced in a `.gitlab-ci.yml` with a multi-line array input followed by a scalar input: completions previously appeared on the array's `- item` lines, one column either side of the keys, and after typing the right indent then moving the cursor left. After the fix, suggestions appear only when the cursor sits in a properly-aligned name slot.

## Breaking Changes
- [x] No breaking changes

## Release Notes Draft
- Fix: GitLab component input suggestions now appear only in a correctly-aligned input slot, not on array values or when the cursor is moved left of the indent.

### DIFF
--- a/src/providers/completionInputContext.ts
+++ b/src/providers/completionInputContext.ts
@@ -37,10 +37,17 @@ function indentOf(line: string): number {
  *
  * @param text - The full YAML document text.
  * @param lineIndex - 0-based line index where the cursor sits.
+ * @param column - 0-based cursor column. When given, the cursor must actually sit at the slot's indent column —
+ *   typing the right indentation then moving the cursor left leaves the spaces in place but no longer offers
+ *   completions. Omit to check the line's text alone (the cursor-agnostic form used by tests).
  * @returns A populated {@link CompletionInputContext} when the cursor is in a parameter-name slot inside the
  *   `inputs:` block of the closest preceding `component:`/`local:` include; `null` otherwise.
  */
-export function findCompletionInputContextAtLine(text: string, lineIndex: number): CompletionInputContext | null {
+export function findCompletionInputContextAtLine(
+  text: string,
+  lineIndex: number,
+  column?: number
+): CompletionInputContext | null {
   const parsed = parseYaml(text);
   if (!isYamlNode(parsed) || !parsed.include) return null;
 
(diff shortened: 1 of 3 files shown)


src/providers/completionInputContext.ts
```diff
@@ -55,12 +62,26 @@ export function findCompletionInputContextAtLine(text: string, lineIndex: number
...
-  // already a complete `key: value`.
+  // parameter-name slot. A name slot must line up at the same column as the existing input keys
+  // (`section.childIndent`) — one column shallower is a sibling of `inputs:`, deeper is a value nested under
+  // another input. When the block has no keys yet, fall back to "deeper than `inputs:`". It must also not be an
+  // array-item line (`- value`, part of a parameter value) nor an already-complete `key: value`.
   const currentLine = lines[lineIndex];
   const currentLineText = currentLine.trim();
+  const currentIndent = indentOf(currentLine);
+  const indentMatches =
+    section.childIndent !== null ? currentIndent === section.childIndent : currentIndent > section.inputsIndent;
```
eFAILution, line 73: Anchoring to `childIndent` is the right call. One coverage gap: the `: currentIndent > section.inputsIndent` fallback (empty `inputs:` block, no keys yet) is the common "adding the first input" path, but it isn't exercised by the new tests — they all have an existing `environment` key, so `childIndent` is always non-null here. Worth a case: empty `inputs:` block, cursor in the slot → detected; cursor left of the slot → null. That locks in the fallback together with the column gate. Non-blocking.

src/providers/completionInputContext.ts
```diff
@@ -55,12 +62,26 @@ export function findCompletionInputContextAtLine(text: string, lineIndex: number
...
+  // When a cursor column is supplied, the cursor must sit at the slot's indent column or within the name being
+  // typed after it — never left of the indent. Typing the right indentation then moving the cursor left leaves
+  // the line's whitespace intact, so the indent check above still matches; but the cursor is no longer in the
+  // name slot, so we must not offer completions there.
+  const cursorAtSlot = column === undefined || column >= currentIndent;
   const isParameterContext =
-    indentOf(currentLine) > section.inputsIndent && (!currentLineText.includes(':') || currentLineText.endsWith(':'));
+    indentMatches &&
+    cursorAtSlot &&
+    !currentLineText.startsWith('- ') &&
```
eFAILution, line 82: Nit: `!startsWith('- ')` + `!== '-'` covers YAML array items well. A no-space `-foo` would slip through, but that parses as the scalar `-foo` (not a sequence item), so it isn't worth guarding — just noting it.

eFAILution (approved):
Approving — clean, well-scoped fix. Slot detection now anchors to the column the existing input keys line up at (`childIndent`), excludes array-item lines, and gates on the cursor column so completions no longer fire when the cursor is moved left along correct indentation.

Reviewed the logic against all 11 pre-existing tests and they stay consistent (e.g. "indented past it" aligns with the key so it still detects; "too shallow" returns null under both the old and new rules), and the only production caller (`completionProvider.ts`) is updated to pass `position.character` — the new `column` param is optional, so nothing else breaks. CI is green. Two minor inline notes below, neither blocking.

PR #226, opened 2026-07-13
fix(parser): resolve include: in multi-document component templates

### LINKED ISSUE #225: bug:  input completion, hover, and validation do nothing on a template with a `spec:` header
### Bug summary

A GitLab CI component template is a **multi-document YAML file**: the `spec:` header is one document and the `include:`/jobs body is a second document, separated by `---`. The extension parses the document with `js-yaml`'s `yaml.load()`, which **throws** on any stream containing more than one document (`expected a single document in the stream, but found more`). The throw is caught and the parse returns `null`, so the completion, hover, and validation providers all see no `include:` and silently do nothing. This affects every real component template that keeps its `spec:` header.

### Steps to reproduce

1. Open a `.gitlab-ci.yml` / template that begins with a `spec:` block, then `---`, then an `include:` block with a `- component:`/`- local:` entry carrying `inputs:` (see the snippet below).
2. Put the cursor on a blank, correctly-indented line inside that include's `inputs:` block.
3. Request input-name completion (auto or Ctrl+Space).

### Expected behavior

- Input-name completions are offered for the include's not-yet-set inputs.
- Hover on an input surfaces its parameter docs, and validation flags unknown/missing inputs — as they already do for a single-document `.gitlab-ci.yml`.

### Actual behavior

- No completions, no hover, no diagnostics. The debug log shows `No inputs parameter-name slot at cursor`.
- Deleting the `spec:` block (and its `---`) collapses the file to a single document and everything starts working — confirming the multi-document stream is the trigger.

_No response_

```yaml
spec:
  inputs:
    environment:
      type: string
      default: staging
    region:
      type: string
    job_name:
      type: string
---
include:
  - local: "templates/deploy.yml"
    inputs:
      job_name: deploy
      # cursor on the blank line below — no completions offered before the fix
```

### Extension logs

```text
N/A
```

### Affected area

Hover provider

### Extension version

0.16.0

### VS Code version

1.128.0

### Node.js version

22.x

### Operating system

macOS

### Pre-submit checks

X-Guardian:
## Summary
- Fixes input completion, hover, and validation doing nothing on a component template that keeps its `spec:` header. Such a template is a **multi-document** YAML stream (`spec:` header + `---` + `include:`/jobs body); `parseYaml` used `js-yaml`'s single-document `load()`, which throws on a multi-document stream, so the parse returned `null` and every provider that reads `include:` bailed. Deleting the `spec:` block was the only workaround.
- Adds `parseYamlDocuments` (parses the whole stream via `loadAll`, returning each mapping document in order) and `findDocumentWith(docs, key)` (selects the document that owns a given top-level key). The three broken providers now select the `include`-bearing document instead of assuming a single document — matching what `localComponentResolver` already does with `loadAll`. Distinct documents are **not** flattened together; each caller picks the document it needs.
- `parseYaml` is unchanged, so the single-document callers (the `quoteYamlIfUnsafe` round-trip probe, the wrapped-include parse in the component browser) keep their exact behaviour.
- Link related issue(s): Fixes #225

## Change Type
- [ ] feat
- [x] fix
- [ ] refactor
- [ ] docs
- [ ] test
- [ ] chore

### User-facing impact
- Completion, hover, and validation now work on `spec:`-header component templates, not only on plain single-document `.gitlab-ci.yml` files.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (local YAML parsing only; no GitLab interaction)

### Affected areas
- [ ] Component Browser
- [x] Hover provider
- [x] Completion provider
- [x] Validation provider
- [ ] Docs only

## What changed by bucket

| Bucket | Files | Approach |
|---|---|---|
| Multi-document parse helpers | [yamlParser.ts](../src/utils/yamlParser.ts) | New `parseYamlDocuments(text, silent)` uses `yaml.loadAll` and returns the stream's mapping documents in document order (non-mapping documents — scalars, sequences, `null` — are dropped); `[]` on empty or unparseable input. New `findDocumentWith(docs, key)` returns the first document whose own `key` is defined, or `null`. `parseYaml` is untouched. |
| Provider parse-selection | [completionInputContext.ts](../src/providers/completionInputContext.ts), [hoverInputContext.ts](../src/providers/hoverInputContext.ts), [validationProvider.ts](../src/providers/validationProvider.ts) | Each path replaces `parseYaml(text)` + `parsed.include` guard with `findDocumentWith(parseYamlDocuments(text, true), 'include')`. In completion, `parseInputDocument` (including its in-progress-name blank-line retry) now returns the `include`-bearing `YamlNode` or `null`. |
| Regression tests | [yamlParser.test.ts](../tests/unit/yamlParser.test.ts), [completionInputContext.test.ts](../tests/unit/completionInputContext.test.ts), [specHeaderCompletion.test.ts](../tests/extension-host/suite/specHeaderCompletion.test.ts), [tests/fixtures/spec-header-completion/](../tests/fixtures/spec-header-completion/) | **Unit:** `parseYamlDocuments`/`findDocumentWith` (multi-doc, single-doc, non-mapping-dropping, unparseable), plus two completion regressions resolving a slot in a `spec:`-header template (blank slot and in-progress name). **Extension-host:** `spec-header-completion` drives VS Code's own completion engine against a consumer file that is itself a `spec:`-header multi-document template; verified load-bearing (fails against the old single-document `parseYaml`, offering only generic word completions). |

## Breaking Changes
- [x] No breaking changes

### DIFF
--- a/src/providers/completionInputContext.ts
+++ b/src/providers/completionInputContext.ts
@@ -5,7 +5,7 @@
  * missing ones).
  */
 
-import { parseYaml, isYamlNode } from '../utils/yamlParser';
+import { parseYaml, parseYamlDocuments, findDocumentWith, isYamlNode, type YamlNode } from '../utils/yamlParser';
 import { findIncludeLine } from '../utils/includeMatcher';
 import type { ComponentParameter } from '../types/git-component';
 
@@ -65,7 +65,7 @@ export function findCompletionInputContextAtLine(
   // exactly while the user is typing the name. Blank the cursor line before parsing: it contributes nothing to the
   // existing-input set (it *is* the slot being typed), and the positional scans below read `lines`, not the parse.
   const parsed = parseInputDocument(text, lines, lineIndex);
-  if (!isYamlNode(parsed) || !parsed.include) return null;
+  if (parsed === null) return null;
 
   const includes = Array.isArray(parsed.include) ? parsed.include : [parsed.include];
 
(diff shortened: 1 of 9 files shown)


eFAILution (approved):
Nice fix — correct root-cause diagnosis (js-yaml `load()` throws on any multi-document stream, so `loadAll` is the right call) and it's cleanly scoped. The two single-document `parseYaml` callers that must stay single-doc (the `quoteYamlIfUnsafe` round-trip probe and the component-browser wrapped-include parse) are correctly left alone, and the completion blank-line retry is preserved. The completion regression test is load-bearing — it genuinely fails against the old single-document parse. Approving.

**Non-blocking nits:**

- **Hover and validation have no end-to-end regression test.** #225 lists all three of completion/hover/validation as broken, but only completion is covered by `specHeaderCompletion.test.ts`. `hoverInputContext.findInputContextAtLine` is pure and unit-testable — a multi-doc case there (and a validation regression on a `spec:`-header consumer) would pin the other two fixes rather than leaving them to rest transitively on the `parseYamlDocuments`/`findDocumentWith` unit tests.
- `findDocumentWith` selects on `doc[key] !== undefined`, so a present-but-null `include:` now picks that document where the old truthy guard rejected it. Harmless downstream (it filters to empty), just a small semantic drift worth a comment.
- `parseYamlDocuments` bypasses the `parseYaml` 5s parse cache. It's a wash on the typing hot path (the text changes every keystroke anyway), and only repeated hovers/validations on an unchanged document lose reuse — fine for typical CI file sizes, worth noting only if large-file latency ever comes up.

PR #228, opened 2026-07-13
Fix(completion): valid local include with no spec is reported as not found

### LINKED ISSUE #227: bug: a valid `local:` include with no `spec:` block is reported as "File not found or unreadable"
### Bug summary

The validation provider warns `Unable to resolve local include '<path>'. File not found or unreadable.` for a `local:` include whose target **exists and is readable** but declares no `spec:` block. `resolveLocalComponent` collapses four distinct outcomes into a single `null` return (unsupported path, no resolution root, unreadable file, and "readable but no `spec:` block"), and the validator treats any `null` as not-found. A legitimate non-parameterised include therefore gets a spurious warning.

### Steps to reproduce

1. Create a plain local include with no `spec:` block (see `no-spec.yml` snippet below).
2. Reference it from a `.gitlab-ci.yml`: `- local: "gitlab/.cache-ci.yml"` (a real, readable file).
3. Open the `.gitlab-ci.yml` and look at the Problems panel / the include line.

### Expected behavior

- No diagnostic. A `local:` include with no `spec:` block is a valid plain include — there are no inputs to validate, so validation should stay silent (the same way it already skips a resolved template that has an empty `spec.inputs`).

### Actual behavior

- A `local-include-not-found` warning (`File not found or unreadable`) is anchored to the include line, even though the file exists and is readable.

_No response_

```yaml
# gitlab/.cache-ci.yml — a valid plain include, no spec: block
.cache-base:
  cache:
    key: { files: [package.json] }
    paths: ["$PNPM_HOME/store"]
    policy: pull



# .gitlab-ci.yml
include:
  - local: "gitlab/.cache-ci.yml"   # exists + readable, but no spec: → spurious "not found" before the fix
```

### Extension logs

```text
N/A
```

### Affected area

Completion provider

### Extension version

0.16.0

### VS Code version

1.128.0

### Node.js version

22.x

### Operating system

macOS

### Pre-submit checks

X-Guardian:
## Summary
- Fixes a valid `local:` include with no `spec:` block being reported as `File not found or unreadable`. `resolveLocalComponent` collapsed four distinct outcomes into a single `null` (unsupported path, no resolution root, unreadable file, and readable-but-no-`spec:`), and the validation provider treated any `null` as not-found — so a legitimate plain include (hidden jobs, cache config) got a spurious warning.
- Adds `resolveLocalIncludeOutcome`, returning a discriminated `LocalIncludeOutcome` (`component` / `skipped` / `unreadable` / `no-spec`). The validator now only emits `local-include-not-found` for `unreadable`; `no-spec` and `skipped` are silent. Unparseable-but-readable YAML is treated as `no-spec` rather than a false not-found.
- `resolveLocalComponent` is kept as a thin wrapper over the new function (returns the component or `null`), so the completion and include-detection callers are unchanged.
- Link related issue(s): Fixes #227

## Change Type
- [ ] feat
- [x] fix
- [ ] refactor
- [ ] docs
- [ ] test
- [ ] chore

### User-facing impact
- A `local:` include of a file with no `spec:` block (a plain include of hidden jobs / cache config) no longer shows a spurious "file not found or unreadable" warning. Genuinely missing/unreadable includes still warn.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (local include resolution only; no GitLab interaction)

### Affected areas
- [ ] Component Browser
- [ ] Hover provider
- [ ] Completion provider
- [x] Validation provider
- [ ] Docs only

## What changed by bucket

| Bucket | Files | Approach |
|---|---|---|
| Resolution outcome | [localComponentResolver.ts](../src/providers/localComponentResolver.ts) | New `LocalIncludeOutcome` union and `resolveLocalIncludeOutcome(localPath, document)`: `skipped` for a glob/`..` path or no resolution root, `unreadable` when the file is missing/unreadable, `no-spec` when it's readable but declares no `spec:` document (or is unparseable YAML), and `component` on success. `resolveLocalComponent` becomes a thin wrapper returning `outcome.component` or `null`, so completion and `detectLocalIncludeComponent` are untouched. |
| Validator branch | [validationProvider.ts](../src/providers/validationProvider.ts) | `validateLocalInclude` switches to `resolveLocalIncludeOutcome`. Only `unreadable` produces the `local-include-not-found` diagnostic; `no-spec`/`skipped` return silently (no inputs to validate). |
| Regression tests | [localInclude.test.ts](../tests/extension-host/suite/localInclude.test.ts) | New extension-host test: a readable `local:` include with no `spec:` block produces **no** diagnostics on its line. Strengthened the existing missing-file test to assert the not-found diagnostic anchors to the `missing-template.yml` line specifically. Both verified load-bearing (fail when the validator warns on `no-spec`). Reuses the existing `local-include/no-spec.yml` fixture. |

## Breaking Changes
- [x] No breaking changes

### DIFF
--- a/src/providers/localComponentResolver.ts
+++ b/src/providers/localComponentResolver.ts
@@ -183,15 +183,91 @@ async function readIncludeTarget(uri: vscode.Uri): Promise<string | null> {
 }
 
 /**
- * Resolve a local include path into a synthetic `Component` so the rest of the providers can treat it identically to
- * a catalog component. Reads the target file (preferring an open editor buffer), parses every YAML document in it,
- * picks the first one declaring `spec`, and maps `spec.inputs` onto `ComponentParameter[]`.
+ * The outcome of resolving a `local:` include, discriminating the reasons resolution can fail to yield a component.
  *
- * Returns `null` for several distinct reasons, all of which the caller treats the same way (no spurious diagnostics):
- *   - the path is a glob or contains `..` (see `isUnsupportedLocalPath`)
- *   - no resolution root is available (no document, no Git repo, no workspace folder)
- *   - the file doesn't exist or can't be read
- *   - the file parses but has no `spec` block (a legitimate plain include, not a parameterised template)
+ * `unreadable` and `no-spec` look identical to callers that only want the component, but they are opposites to the
+ * validator: `unreadable` is a real problem (the included file is missing or can't be read) worth a diagnostic,
+ * whereas `no-spec` is a legitimate plain include (hidden jobs, `.pre`/`.post`, cache config) with nothing to
+ * validate and must stay silent. Collapsing both to `null` is what made valid non-parameterised includes report a
+ * spurious "file not found" warning.
+ */
+export type LocalIncludeOutcome =
+  | { kind: 'component'; component: Component }
+  /** Path is a glob/`..`, or no resolution root is available — not something to diagnose. */
+  | { kind: 'skipped' }
+  /** The include target is missing or unreadable — the validator surfaces this as a diagnostic. */
+  | { kind: 'unreadable' }
+  /** The target was read and parsed but declares no `spec:` document — a valid plain include, nothing to validate. */
+  | { kind: 'no-spec' };
+
+/**
+ * Resolve a local include path into a {@link LocalIncludeOutcome}. Reads the target file (preferring an open editor
+ * buffer), parses every YAML document in it, picks the first one declaring `spec`, and maps `spec.inputs` onto
+ * `ComponentParameter[]`.
+ *
+ * @param localPath  The raw path string from a parsed `include: - local:` entry.
+ * @param document   The document making the include. Used to determine the resolution root and the open-buffer fast path.
+ * @returns          The resolution outcome — a `component` on success, or one of `skipped`/`unreadable`/`no-spec`.
+ */
+export async function resolveLocalIncludeOutcome(
+  localPath: string,
+  document?: vscode.TextDocument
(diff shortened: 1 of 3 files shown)


eFAILution (approved):
Approving — the discriminated `LocalIncludeOutcome` cleanly separates the four cases, the validator now emits `local-include-not-found` only for `unreadable`, and the `resolveLocalComponent` wrapper keeps the completion and include-detection callers byte-for-byte compatible (both only test truthiness). The no-spec regression test genuinely fails against the old code.

**One deliberate behavior change I'd like to confirm before this is final (non-blocking, but please address in a follow-up):**

- **Unparseable-but-readable YAML is now silently reclassified as `no-spec`** (`localComponentResolver.ts` ~L74-75). A genuinely malformed include template — including a file that *has* a valid `spec:` document but a broken second document after `---`, which makes `yaml.loadAll` throw — now produces zero diagnostics instead of surfacing anything. That's defensible (YAML validity is arguably the YAML language server's / GitLab lint's job, and a false "not found" on a readable file is worse than silence), but it masks a real class of user error and isn't called out in the description. Either add a distinct `unparseable` outcome with a correctly-worded hint ("local include exists but couldn't be parsed as YAML"), or document the decision explicitly. It also ships untested — please add a malformed-YAML fixture asserting no false not-found.

**Minor:** the refactor also makes the "no resolution root" case (a `.gitlab-ci.yml` open with no workspace folder) silent instead of emitting a spurious not-found on every `- local:` line — a nice side-effect bug fix worth a line in the description.

PR #230, opened 2026-07-13
fix(browser): hover-fetched components create a phantom top-level instance group in the browser

### LINKED ISSUE #229: bug: hover-fetched components create a phantom top-level instance group in the browser
### Bug summary

Hover-fetching a tag-pinned component that belongs to a configured monorepo source creates a phantom top-level group in the Component Browser, named after the GitLab instance, containing a duplicate of the component. The component then appears twice — once correctly under its configured source, once wrongly under the instance group.

### Steps to reproduce

1. Configure a single monorepo source:
   ```jsonc
   "gitlabComponentHelper.componentSources": [
     {
       "name": "my-shared-config",
       "path": "my-group/my-subgroup/my-shared-config",
       "gitlabInstance": "gitlab.com",
       "type": "project",
       "tagPattern": "{name}-{version}"
     }
   ]
   ```
2. Run "GitLab CI: Browse Components" — confirm the browser shows the expected single source group (`my-shared-config`, with its components).
3. Open a `.gitlab-ci.yml` and hover a tag-pinned component link belonging to that source (see snippet below).
4. Re-open / refresh the Component Browser.
5. Observe a second, phantom top-level group named after the GitLab instance (e.g. `gitlab.com`) containing a duplicate of the hovered component.

### Expected behavior

A hover-fetched component that belongs to a configured source merges into that source's group, carrying the source's display name and `tagPattern`. No phantom instance-level group appears.

### Actual behavior

A second, phantom top-level group named after the GitLab instance (e.g. `gitlab.com`) appears, containing a duplicate entry for the hovered component with:

- no description ("Component/Project does not have a description"),
- the raw tag (e.g. `my-component-1`) as its version instead of a scoped/stripped value (e.g. `1`),
- no version dropdown.

The component is listed twice — once correctly under its configured source group, once wrongly under the instance group.

_No response_

```yaml
include:
  - component: $CI_SERVER_FQDN/my-group/my-subgroup/my-shared-config/my-component@my-component-1
```

### Extension logs

```text
N/A
```

### Affected area

Component Browser

### Extension version

0.16.0

### VS Code version

1.128.0

### Node.js version

22.x

### Operating system

macOS

### Pre-submit checks

X-Guardian:
## Summary
- Fixes a **phantom top-level source group** in the Component Browser. Hovering a tag-pinned component that isn't already cached triggers a dynamic fetch; the fetched cache entry was labelled with a synthesized `source` of `` `${gitlabInstance}/${projectPath}` `` (e.g. `gitlab.com/yu-life/infrastructure/yulife-devops-shared-config`) instead of the configured source's display name. Since the browser groups by `source.split('/')[0]`, this created a duplicate `gitlab.com` group alongside the real source, with the component showing no description, a raw unstripped tag (`markdown-lint-1`), and no version dropdown.
- **New pure helper** (`src/services/cache/sourceReconciliation.ts`): `reconcileComponentSource(component, sources)` matches a fetched component against the configured sources by `gitlabInstance` + `sourcePath` and, on a match, adopts the source's display `name` and fills in its `tagPattern` (a `tagPattern` already on the component wins). No match returns the component unchanged. Instance comparison normalizes a leading `https?://` and defaults an unset `gitlabInstance` to `gitlab.com`.
- **Wired into the single cache-write chokepoint**: `ComponentCacheManager.addDynamicComponent` reads `componentSources` and runs the reconciliation before dedup, so **both** dynamic-fetch call sites in `componentDetector.ts` (specific-version fetch and catalog-fallback fetch) are covered, plus any future callers.
- Adds `tagPattern?` to the `ComponentSource` type (`src/types/api.ts`) so the config shape is typed at the read site.
- Link related issue(s): Fixes #229 

## Change Type
- [ ] feat
- [x] fix
- [ ] refactor
- [ ] docs
- [x] test
- [ ] chore

### User-facing impact
- After hovering a tag-pinned component belonging to a configured source, the entry now merges into that source's group instead of spawning a duplicate `gitlab.com` top-level group.
- Because the merged entry inherits the source's `tagPattern`, its version discovery scopes/strips correctly (e.g. `1` rather than the raw `markdown-lint-1`).
- No change for hovers on projects outside any configured source, or for ordinary single-component sources.
- **Existing caches**: the fix prevents new phantom entries but does not retroactively relabel an entry already persisted from a previous session. `CURRENT_CACHE_VERSION` is not bumped, so users with the phantom group must **Reset Cache** (or Refresh) once. (Open question for review — see Risk/Rollback.)

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (matching logic is instance-independent)

### Affected areas
- [x] Component Browser
- [x] Hover provider
- [ ] Completion provider
- [ ] Validation provider
- [ ] Docs only

### Local checks
- [x] `npm run compile`
- [ ] Manual verification in VS Code Extension Host

New unit tests in `tests/unit/sourceReconciliation.test.ts` cover: name adoption on an instance+path match; `tagPattern` fill vs. preserve precedence; no-match passthrough; cross-instance non-match; `https://`-prefix normalization; and default-to-`gitlab.com` when a source omits `gitlabInstance`. Full suite: 336 passing.

## Breaking Changes
- [x] No breaking changes

The added `tagPattern?` field on `ComponentSource` is additive and optional. No settings or persisted-cache shape changes.

## Risk and Rollback
- Main risks:
  - Low. The reconciliation only rewrites `source`/`tagPattern` on entries that positively match a configured source; unmatched entries are untouched.
- Stale caches: existing persisted phantom entries are not migrated. Options for review: (a) accept a one-time manual **Reset Cache**, or (b) bump `CURRENT_CACHE_VERSION` so old caches self-invalidate on upgrade. This PR takes (a); happy to switch to (b) if preferred.
- Rollback strategy: revert the PR. No data migration; the settings/cache shapes are unchanged.

## Release Notes Draft
- Fixed hover-fetched components appearing as a duplicate top-level `gitlab.com` group in the Component Browser; they now merge into their configured source and inherit its tag pattern.

## Checklist
- [ ] Added/updated docs for behavior or settings changes

### DIFF
--- a/src/services/cache/componentCacheManager.ts
+++ b/src/services/cache/componentCacheManager.ts
@@ -3,6 +3,8 @@ import { getComponentService } from '../component';
 import { Logger } from '../../utils/logger';
 import { getPerformanceMonitor } from '../../utils/performanceMonitor';
 import { CachedComponent, PersistentCacheData } from '../../types/cache';
+import { ComponentSource } from '../../types/api';
+import { reconcileComponentSource } from './sourceReconciliation';
 import { ProjectCache } from './projectCache';
 import { VersionCache } from './versionCache';
 import { GroupCache } from './groupCache';
@@ -115,6 +117,13 @@ export class ComponentCacheManager {
    */
   public addDynamicComponent(component: CachedComponent): void {
     try {
+      // Dynamic fetches synthesize `source` as `instance/path`; adopt the configured source's display name (and tag
+      // template) so the entry merges into its source group instead of spawning a phantom top-level instance node.
+      const sources = vscode.workspace
+        .getConfiguration('gitlabComponentHelper')
+        .get<ComponentSource[]>('componentSources', []);
+      component = reconcileComponentSource(component, sources);
+
       // Check if component already exists (avoid duplicates)
       const existingIndex = this.components.findIndex(
         comp =>
--- a/src/services/cache/sourceReconciliation.ts
+++ b/src/services/cache/sourceReconciliation.ts
@@ -1,0 +1,39 @@
+import { CachedComponent } from '../../types/cache';
+import { ComponentSource } from '../../types/api';
+
+const DEFAULT_GITLAB_INSTANCE = 'gitlab.com';
+
+const normalizeInstance = (instance: string): string => instance.replace(/^https?:\/\//, '');
+
+/**
+ * When a dynamically fetched component belongs to a configured source, return it with that source's display `name`
+ * adopted (and `tagPattern` filled in if absent). Dynamic fetches synthesize `source` as `instance/path`, which would
+ * otherwise group the component under a phantom top-level instance node in the browser — separate from the configured
+ * source's own components — and leave version discovery without the tag template.
+ *
+ * Matching is by `gitlabInstance` + `sourcePath`. No match returns the component unchanged (an ad-hoc hover on a
+ * project outside any configured source).
+ *
+ * @param component The dynamically fetched cache entry, with `source` synthesized as `instance/path`.
+ * @param sources   The configured component sources from settings to match against.
+ * @returns         The component with `source`/`tagPattern` adopted from the matching source, or unchanged if none match.
+ */
+export function reconcileComponentSource(
+  component: CachedComponent,
+  sources: ComponentSource[],
+): CachedComponent {
+  const match = sources.find(source => {
+    const sourceInstance = normalizeInstance(source.gitlabInstance || DEFAULT_GITLAB_INSTANCE);
+    return sourceInstance === component.gitlabInstance && source.path === component.sourcePath;
+  });
+
+  if (!match) {
+    return component;
+  }
+
+  return {
+    ...component,
+    source: match.name,
+    tagPattern: component.tagPattern ?? match.tagPattern,
+  };
+}
--- a/src/types/api.ts
+++ b/src/types/api.ts
@@ -128,6 +128,7 @@ export interface ComponentSource {
   path: string;
   gitlabInstance?: string;
   type?: 'project' | 'group';
+  tagPattern?: string;
 }
 
 /**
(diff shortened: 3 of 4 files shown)


eFAILution (approved):
Approving — clean, correctly-scoped fix for the reported `type: project` repro. The reconciliation sits at the single `addDynamicComponent` chokepoint, so all the dynamic-fetch call sites are covered, and since the dedup key doesn't include `source`, rewriting it can't cause a wrongful merge. The pure helper is well unit-tested.

**Non-blocking, but two gaps worth a follow-up:**

- **`type: group` sources still spawn the phantom.** Reconciliation matches `source.path === component.sourcePath`, but group members are stored with `sourcePath = full path_with_namespace` while the configured `path` is the group path — so a hovered group member never matches and the `gitlab.com` top-level group persists for group sources. Same bug class as #229, just half-covered. Consider a `component.sourcePath.startsWith(source.path + '/')` match for `type === 'group'`, adopting `${source.name}/${lastSegment}` to line up with `groupCache`'s shape.
- **Instance normalization only strips a leading `https?://`.** Trailing slash (`https://gitlab.com/`), case (`GitLab.com`), uppercase scheme, and port all fail to match — each fails *safe* (the phantom persists, never a wrong match), but the user sees the bug with a valid-looking config. `.replace(/^https?:\/\//i, '').replace(/\/+$/, '').toLowerCase()` covers them.

**Test gap:** the pure helper is covered, but its wiring into `addDynamicComponent` and the actual user-visible "phantom group is gone / duplicate merges" behavior through `transformCachedComponentsToGroups` aren't. A transform-level regression test would pin #229's actual symptom.

**Stale-cache limitation** (no `CURRENT_CACHE_VERSION` bump) is acceptable — the on-disk shape is unchanged and it self-heals on the next full refresh or re-hover, so option (a) is fine.

PR #232, opened 2026-07-13
fix(links): show component links on first view

### LINKED ISSUE #231: bug: component links are absent on first view until the file is edited
### Bug summary

When a GitLab CI file is first opened, the clickable `component:` links are missing. They only appear after the first edit to the file, even when the component cache is already populated.

### Steps to reproduce

1. Ensure the component cache is populated (open the extension once so a source is discovered and cached).
2. Open a `.gitlab-ci.yml` that references a cached component (see snippet below).
3. Observe the `component:` URL — it is plain text, not a clickable link.
4. Type any character (and undo it) to force a document change.

### Expected behavior

The `component:` URL is a clickable link as soon as the file is rendered, pointing at the component template on GitLab.

### Actual behavior

No link on first view. The link only appears after the first document change forces VS Code to re-request links from the provider.

_No response_

```yaml
include:
  - component: $CI_SERVER_FQDN/group/project/my-component@1.0.0
    inputs:
      environment: staging
```

### Extension logs

```text
N/A
```

### Affected area

Other

### Extension version

0.16.0

### VS Code version

1.128.0

### Node.js version

22.x

### Operating system

macOS

### Pre-submit checks

X-Guardian:
## Summary
- Fixes `component:` links being absent when a GitLab CI file is first viewed — they previously only appeared after the first edit, even with a populated cache.
- The document link provider resolves links asynchronously from the component cache but never implemented VS Code's optional `onDidChangeLinks` event. VS Code requests links once per document render, caches the result, and re-requests only on a document change or when `onDidChangeLinks` fires. With no change event, the first-render result stuck until the first edit forced a re-request.
- Adds an `onDidChangeComponents` event to the cache manager, fired whenever the in-memory component list is populated/changed; the provider subscribes and re-fires its own `onDidChangeLinks` so VS Code re-requests once the cache is ready.
- The provider also fires an initial refresh *after* registration (a fire before VS Code has subscribed is discarded), so an editor already open at activation re-requests against the current cache.
- The cache event is coalesced onto a microtask and suppressed on no-op re-adds, so a validation pass (which re-adds every include's component on each open/change/save) triggers at most one re-request instead of one per component.
- Link related issue(s): Fixes #231

## Change Type
- [ ] feat
- [x] fix
- [ ] refactor
- [ ] docs
- [ ] test

### User-facing impact
- Clickable component links now appear as soon as a GitLab CI file is rendered, without needing to edit the file first.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (local link-refresh scheduling only; no GitLab interaction)

### Affected areas
- [ ] Component Browser
- [ ] Hover provider
- [ ] Completion provider
- [ ] Validation provider
- [ ] Docs only

## What changed by bucket

| Bucket | Files | Approach |
|---|---|---|
| Cache change event | [componentCacheManager.ts](../src/services/cache/componentCacheManager.ts) | Added an `onDidChangeComponents` `EventEmitter` and a private `notifyComponentsChanged()` helper, fired at every point the in-memory list is populated/changed: the valid-disk-cache early return in `initializeCache` (the first-view "already populated" path), the end of `refreshComponentsInternal`, and `addDynamicComponent`. `notifyComponentsChanged()` coalesces a burst of mutations into one emission via `queueMicrotask`; `addDynamicComponent` skips the fire when nothing link-relevant changed. Manager is now `Disposable` — disposes the emitter and the (previously un-disposed) config-change subscription. |
| No-op guard | [cachedComponentEquality.ts](../src/utils/cachedComponentEquality.ts) | `sameCachedComponent` — pure comparison of the render-relevant fields of two already-identity-matched entries. Extracted to a vscode-free util so it is unit-testable. |
| Link refresh | [documentLinkProvider.ts](../src/providers/documentLinkProvider.ts) | Provider now exposes `onDidChangeLinks`, subscribes to `onDidChangeComponents` to re-fire it, and exposes `refresh()` (called after registration) so an already-open editor re-requests against the current cache. Provider is now `Disposable` and cleans up its subscription and emitter. |
| Registration + initial fire | [extension.ts](../src/extension.ts) | Register the provider and cache manager in `context.subscriptions` so their `dispose()` runs on deactivation. Call `documentLinkProvider.refresh()` via `queueMicrotask` *after* `registerDocumentLinkProvider`, so the initial fire lands once VS Code has subscribed rather than being discarded. |
| Unit test | [cachedComponentEquality.test.ts](../tests/unit/cachedComponentEquality.test.ts) | Covers the no-op guard: an identical re-add is equal (suppresses the fire); a change to `templatePath`/`url`/`source`/`tagPattern`/`resolvedSha`/`availableVersions` is a real change; non-link fields are ignored. |

### Local checks
- [x] `npm run compile` (`tsc --noEmit` clean)
- [x] `npm test` (339 passing, incl. the new no-op-guard suite)
- [x] `npm run test:extension-host` (20 passing)
- [x] `npm run lint` clean

### Manual test notes
- Reproduced with a populated cache: opening a `.gitlab-ci.yml` showed no link on the `component:` line until the first edit. After the fix, the link appears on first render.

## Breaking Changes
- [x] No breaking changes

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -116,12 +116,14 @@ export function activate(context: vscode.ExtensionContext) {
     // Register document link provider so `component:` URLs (including those using GitLab variables like
     // $CI_SERVER_FQDN) become clickable and point at the GitLab project tree at the requested ref.
     logger.debug('[Extension] Registering document link provider...', 'Extension');
+    const documentLinkProvider = new ComponentDocumentLinkProvider();
     context.subscriptions.push(
+      documentLinkProvider,
       vscode.languages.registerDocumentLinkProvider(
         [
           { language: 'yaml' }
         ],
-        new ComponentDocumentLinkProvider()
+        documentLinkProvider
       )
     );
 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -70,6 +70,7 @@ export function activate(context: vscode.ExtensionContext) {
     let cacheManager: ComponentCacheManager;
     try {
       cacheManager = getComponentCacheManager(context);
+      context.subscriptions.push(cacheManager);
       logger.info(`[Extension] Component cache manager initialized successfully`, 'Extension');
     } catch (cacheError) {
       logger.error(`[Extension] ERROR initializing cache manager: ${cacheError}`, 'Extension');
@@ -126,6 +127,8 @@ export function activate(context: vscode.ExtensionContext) {
         documentLinkProvider
       )
     );
+    // Fire the initial refresh after registration.
+    queueMicrotask(() => documentLinkProvider.refresh());
 
     // Register command to add project/group token
     logger.debug('[Extension] Registering addProjectToken command...', 'Extension');
--- a/src/providers/documentLinkProvider.ts
+++ b/src/providers/documentLinkProvider.ts
@@ -21,11 +21,14 @@ export class ComponentDocumentLinkProvider implements vscode.DocumentLinkProvide
   private readonly cacheSubscription: vscode.Disposable;
 
   constructor() {
-    // Re-request links whenever the cache changes. Fire once up front too, so an editor already open at registration
-    // re-requests against the current cache rather than keeping the (possibly empty) result from its first render.
     this.cacheSubscription = getComponentCacheManager().onDidChangeComponents(() => {
       this._onDidChangeLinks.fire();
     });
+  }
+
+  // Re-request links against the current cache. Call after registration so an editor already open at activation
+  // re-requests instead of keeping the (possibly empty) result from its first render.
+  public refresh(): void {
     this._onDidChangeLinks.fire();
   }
 
--- a/src/providers/documentLinkProvider.ts
+++ b/src/providers/documentLinkProvider.ts
@@ -9,9 +9,30 @@ import { isGitLabCIFile } from '../utils/gitlabCiFileMatcher';
 
 const COMPONENT_LINE = /^(\s*-?\s*component:\s*)(\S+)/;
 
-export class ComponentDocumentLinkProvider implements vscode.DocumentLinkProvider {
+export class ComponentDocumentLinkProvider implements vscode.DocumentLinkProvider, vscode.Disposable {
   private logger = Logger.getInstance();
   private urlParser = new UrlParser();
+
+  // VS Code requests links once per document render and caches the result, re-asking only when the document changes or
+  // this event fires. Without it, links resolved from the async cache never appear on first view — they only show up
+  // after the first edit. We fire it when the component cache is populated/updated so VS Code re-requests then.
+  private readonly _onDidChangeLinks = new vscode.EventEmitter<void>();
+  public readonly onDidChangeLinks = this._onDidChangeLinks.event;
+  private readonly cacheSubscription: vscode.Disposable;
+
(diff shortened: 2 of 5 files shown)


eFAILution (changes requested):
The fix does resolve #231 today, but it works by timing coincidence rather than the mechanism documented in the code, and it ships with an event-storm and no tests — requesting changes.

**1. The "fire once on construction" safety net is dead code.** The constructor fires `onDidChangeLinks` before `registerDocumentLinkProvider` runs, so VS Code hasn't subscribed yet and `EventEmitter.fire()` doesn't replay to late subscribers — the initial fire is discarded. The fix actually works only because `activate()` has no `await` between the cache manager's creation and provider registration, so `initializeCache`'s async continuation fires `notifyComponentsChanged` as a microtask *after* registration. Add one `await` before provider registration someday and #231 regresses silently. Please fire the initial event *after* registration (or `queueMicrotask(() => this._onDidChangeLinks.fire())`) and correct the inline comment, which currently describes behavior that doesn't happen.

**2. Event storm.** `notifyComponentsChanged` fires on every `addDynamicComponent`, including the no-op "update with identical data" path, and that runs inside a per-include loop during validation (which itself runs on open/change/save). A file with N components can fire `onDidChangeLinks` up to N times per pass, each forcing a full `provideDocumentLinks` re-scan — roughly O(N²) on large CI files. Fire only when the component set actually changed, and/or debounce/coalesce the emitter so a burst of adds yields one re-request. (Not an infinite loop — the link provider doesn't call back into `addDynamicComponent` — but real thrash.)

**3. No automated tests, and the suite/lint/extension-host weren't run** (per the PR checklist — only `tsc`). Both are plain `vscode.EventEmitter`s, so a unit test that populating the cache fires `onDidChangeComponents` and that the provider re-fires `onDidChangeLinks` is low-cost. At minimum, please run `npm test` + lint before this merges.

**Minor:** the cache manager's `_onDidChangeComponents` emitter is never disposed (it matches a pre-existing un-disposed subscription in the same constructor, so impact is low). The provider side is handled correctly — `dispose()` releases both the subscription and the emitter, and it's registered in `context.subscriptions`.

eFAILution (approved):
Re-reviewed the "Fix review comments" update against the branch — all four findings are resolved. Approving.

- **Initial fire timing:** moved out of the constructor into `refresh()` and invoked via `queueMicrotask(...)` *after* `registerDocumentLinkProvider`, so it lands once VS Code has subscribed; the misleading comment is corrected. #231 stays fixed.
- **Event storm:** gone — the emit is now gated on an actual change (`sameCachedComponent`) and a synchronous burst is coalesced into a single emission. First-populate still fires unconditionally, so links appear on first render.
- **Disposal:** `onDidChangeComponents` is disposed now, and you also captured the previously-leaked `onDidChangeConfiguration` listener — nice catch.
- **Tests:** `cachedComponentEquality` has a load-bearing unit test covering equal→true and each differing field→false.

Non-blocking follow-up (not for this PR): the event wiring itself — the change-gating, the microtask coalescing, and the emitter→provider re-fire — is still only exercised indirectly. A small extension-host or mocked-`EventEmitter` test would lock in the first-view behavior against future refactors. Happy to merge without it.

PR #272, opened 2026-09-08
chore(parser): clarify referenceTag identify option purpose (#270)

### LINKED ISSUE #270: chore(parser): `identify: () => false` on referenceTag is dead code and its comment is wrong
Follow-up from #263. Cosmetic, but the comment is actively misleading.

## Problem

`referenceTag` in `src/utils/yamlParser.ts` passes `identify: () => false`, with this comment:

> Load-only. `identify` selects the tag when *dumping*; returning false stops it claiming the plain arrays this constructs, which would re-emit unrelated sequences as `!reference`.

The behaviour it describes cannot happen, so the option is dead code and the comment will mislead the next reader into preserving it.

## Evidence

Verified against `js-yaml@5.4.1`:

- `defineSequenceTag` already defaults `identify` to `null` when the option is omitted.
- The dumper only ever picks a tag via `if (tag.identify && tag.identify(object))`, so a `null` identify is never selected.
- `yaml.dump()` output is byte-identical with and without the option.

On top of that, nothing under `src/` calls `yaml.dump` at all, and `dump`'s default schema is not `GITLAB_CI_SCHEMA`, so there is no code path where this could matter today.

## Suggested fix

Either drop the option and the comment, or keep the option and rewrite the comment to say plainly that it is belt-and-braces for a future dump call site rather than protection against something js-yaml would otherwise do.

Cid-oe:
## Summary
Resolves #270.

Clarifies the comment on `referenceTag`'s `identify: () => false` in `src/utils/yamlParser.ts`.

## Changes
- Updated the comment on `identify` to accurately reflect the TypeScript `SequenceTagOptions` type contract (where `identify: (data: any) => boolean` is required by `js-yaml`) rather than describing a misleading claim about js-yaml's dumper behavior.
- Retained `identify: () => false` to satisfy the TypeScript interface while keeping the tag strictly load-only.

## Validation
- `npm run check-types` (TypeScript check passes with zero errors).
- Unit tests pass (387 passing tests).
- `git diff --check` passes cleanly.

### DIFF
--- a/src/utils/yamlParser.ts
+++ b/src/utils/yamlParser.ts
@@ -15,8 +15,8 @@ const referenceTag = yaml.defineSequenceTag<unknown[]>('!reference', {
   addItem: (carrier, item) => {
     carrier.push(item);
   },
-  // Load-only. `identify` selects the tag when *dumping*; returning false stops it claiming the plain arrays this
-  // constructs, which would re-emit unrelated sequences as `!reference`.
+  // Load-only sequence tag. `identify` is required by js-yaml's SequenceTagOptions; returning false ensures this
+  // tag is never selected when dumping JavaScript arrays.
   identify: () => false,
 });
 


eFAILution:
Thanks @Cid-oe. Closing as superseded by #273, which replaced `referenceTag` with catch-all local tags and removed the block this edits, so there is no longer a comment here to correct.

Worth recording though: **you were right and the issue was wrong.** #270 claimed the `identify` option was dead code that could simply be deleted. It is a no-op at runtime, but js-yaml's `SequenceTagOptions` types it as required, so removing it fails the build with TS2345. Your comment saying it is required is the accurate one, and #273 keeps the option for the same reason. Sorry for sending you at a bad premise.

PR #275, opened 2026-09-08
refactor(webview): serve loading-view CSS from linted external file under CSP

X-Guardian:
## Summary
- Establishes a reusable pattern for serving webview assets (CSS now, client scripts in a follow-up) from real files instead of strings baked into TypeScript template literals, so they get linting, editor support, and a Content-Security-Policy. Piloted on the Component Browser's loading view as the smallest, lowest-risk surface.
- The loading view's CSS moves from an inline `<style>` block into `src/webview/styles/loading.css`, built by a second esbuild context and loaded via a CSP'd `<link>` resolved through `webview.asWebviewUri`. A shared `src/webview/webviewHtml.ts` helper provides the nonce / CSP / asset-URI plumbing every webview will reuse.
- Adds stylelint for the new CSS and wires both eslint and stylelint to emit GitHub Actions annotations in CI.
- **Step one of six.** This PR puts 1 of the 6 webview documents in `componentBrowserProvider.ts` under a CSP — deliberately the one with no scripts and no remote data. The other five still emit inline `<script>` and interpolate component names, descriptions and tag lists fetched from GitLab; those are the documents a CSP would actually protect, and they follow in the sequence below.

## Change Type
- [ ] feat
- [ ] fix
- [x] refactor
- [ ] docs
- [ ] test
- [x] chore (lint tooling)

### User-facing impact
- No behavior change. The loading spinner looks and behaves exactly as before; only how its CSS is delivered changed (external file under a CSP rather than an inline `<style>`).

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (no GitLab interaction; webview rendering only)

### Affected areas
- [x] Component Browser (loading view rendering only)
- [ ] Hover provider
- [ ] Completion provider
- [ ] Validation provider
- [ ] Docs only

## What changed by bucket

| Bucket | Files | Approach |
|---|---|---|
| Webview HTML helper | `src/webview/webviewHtml.ts` | New shared helpers: `createNonce()` (CSP `script-src 'nonce-…'`), `assetUri()` (resolves an `out/webview` asset to a `vscode-webview://` URL via `webview.asWebviewUri`), and `cspMetaTag()` (builds the `<meta>` CSP — `default-src 'none'` with explicit `style-src`/`img-src`/`font-src`/`script-src`). No `'unsafe-inline'`: styles and scripts must come from files under the webview's own origin, so a document that inlines a style is blocked rather than silently permitted. The nonce is scaffolding for the follow-up script extraction; the loading view has no script yet. |
| CSS extraction | `src/webview/styles/loading.css` | The loading/spinner styles, lifted verbatim from the inline `<style>` in `getLoadingHtml`. Theme colours arrive as `var(--vscode-*)` custom properties, which an external stylesheet reads without needing inline style capability. |
| Loading view render | `src/providers/componentBrowserProvider.ts` | `getLoadingHtml` now takes the `webview`, emits a CSP `<meta>` and a `<link>` to the external stylesheet instead of an inline `<style>`. Panel `localResourceRoots` gains `out/webview` so the asset is loadable, and drops a stale `media` entry — no such directory exists in the repo (icons live in `images/`), and this was its only reference. |
| Build: asset discovery | `esbuild.js` | Split into two contexts: the existing Node extension bundle, plus a browser-target webview context emitting to `out/webview/`. Entry points are **discovered** from `styles/` and `client/` rather than hand-listed — a stylesheet missing from a list would still lint, typecheck and build clean, with the `<link>` only 404ing once packaged. `outbase: 'src/webview'` preserves the subfolders so built paths match what `assetUri` requests, and `outdir` is cleared each build so a renamed or deleted asset can't leave a stale copy that still resolves. `target: es2020` and `format: iife` pin the output for the Electron renderer instead of inheriting esbuild's `esnext` default. |
| Build: watch markers | `esbuild.js` | The problem-matcher plugin was registered on **both** contexts, so every build printed `[watch] build started`/`finished` twice. Since VS Code's background matcher treats the first `finished` as "task ready", it would call the task done while the second build was still running. The markers now come from the extension context only; the webview context uses a label-only error reporter. (Also guards `location`, which is optional on esbuild errors — the old reporter would throw on an error without one.) |
| CSS linting | `stylelint.config.mjs`, `package.json` | Adds stylelint + `stylelint-config-standard` (minimal config, no rule overrides), covering `src/webview/**/*.css`. |
| Lint composition | `package.json` | `lint` and `lint:ci` each split into per-linter scripts composed with `npm-run-all -c` (already a devDependency). Two fixes: the previous `&&` short-circuited, so a PR with both a JS and a CSS error only ever got the JS annotations and needed a second round trip; and the stylelint formatter was referenced by a path into `node_modules` internals, which holds only for npm's flat layout. It is now referenced by package name and resolves through the package's `exports` map. |
| Docs | `.ai/architecture.yaml`, `.ai/workflows.yaml` | `AGENTS.md` designates `.ai/` as the specification, and it had drifted: the architecture file documented a `src/templates/` directory that no longer exists and never mentioned `src/webview/`. Replaced with an accurate `src/webview/` entry; noted `lint:ci` alongside `lint` in workflows. |

### Local checks
- [x] `npm run compile`
- [x] `npm test` — 387 passing; `npm run lint` clean
- [x] Extension-host suite — 23 passing

### Manual test notes
- Ran **GitLab CI: Browse Components** in the Extension Development Host; the loading view renders the spinning loader and "Loading GitLab CI/CD components…" exactly as before.
- Confirmed via webview devtools that `styles/loading.css` loads (200, `vscode-webview://` origin) with no CSP violations in the console.
- Verified the asset wiring is load-bearing: temporarily renaming `webviewHtml.ts` fails the build (`Could not resolve "../webview/webviewHtml"`); moving the built `loading.css` aside removes the spinner (confirming it comes from the external stylesheet, not a default).
- **Asset discovery**: added a second stylesheet with no config change and confirmed it was emitted; deleted it and confirmed the stale output was cleared rather than left resolvable.
- **Watch markers**: counted `[watch] build started` in a full build — 1, previously 2.
- **Lint aggregation**: introduced a JS error and a CSS error simultaneously; both are now reported in a single run and the script exits non-zero. Under the previous `&&` only the JS error appeared.
- Both `.ai/` files still parse as YAML after editing.

## Breaking Changes
- [x] No breaking changes

- No visual change — the loading view is pixel-identical. (Spinner + "Loading GitLab CI/CD components…".)

## Risk and Rollback
- Main risks: low. Scoped to the loading view's CSS delivery; an incorrect asset path or CSP would manifest as an unstyled loading view (no spinner) with no effect on component data, hover, completion, or validation. The packaged asset must ship — `out/webview/**` is not excluded by `.vscodeignore` (confirmed via `vsce ls`) and the panel's `localResourceRoots` includes it.
- The build changes affect every build, not just the webview one: entry-point discovery, the `outdir` clean, and the marker change all sit in `esbuild.js`. A mistake there is loud rather than silent — the build fails or emits nothing — and `npm run compile` plus the packaged-asset check above cover it.
- Rollback strategy: revert the PR; `getLoadingHtml` returns to its inline `<style>`. No persisted state, settings, or cache shape changes.

## Follow-ups
Deliberately out of scope here, tracked for the next PRs in the sequence:
- Extend the CSP to the remaining five webview documents. Each needs its inline `<script>` extracted and its inline event handlers (33 across the five) converted to delegated listeners first, since a nonce CSP blocks `onclick=` attributes.
- Make `webviewHtml.ts` unit-testable. It imports `vscode` at module scope, unlike its `vscode`-free siblings `inlineMarkdown.ts` and `scriptData.ts`, so a unit test importing it would abort the whole mocha run. Taking `cspMetaTag(cspSource: string, nonce: string)` leaves `assetUri` as the only function needing `vscode`.
- `createNonce` uses `byte % 62` over bytes 0–255, so the first 8 characters of the alphabet are ~25% over-represented. Not exploitable at 32 characters, but the doc comment makes a cryptographic claim; rejection-sampling or a 16/64-character alphabet fixes it.
- `assetUri` keeps `..` segments after `filter(Boolean)`, so a computed path could escape `out/webview`. Unreachable from today's single literal call site.
- `getLoadingHtml` embeds a fresh nonce per call, so the two back-to-back loading assignments in the cache-reset path now rebuild the document and visibly restart the spinner. Hoisting the nonce restores a stable string.

## Release Notes Draft
- Internal: webview CSS is now served from linted external files under a Content-Security-Policy (loading view); no user-facing change.

## Checklist
- [x] Added/updated docs for behavior or settings changes (this PR description and `.ai/`; no user-facing settings changed)
- [ ] Added/updated tests for new behavior (no behavior change; existing suite green. `webviewHtml.ts` is not unit-testable until the `vscode` import is split out — see Follow-ups)


eFAILution (changes requested):
Thanks for this — the direction is right and nothing here regresses. There was no CSP anywhere before, the build wiring works (I confirmed `out/webview/styles/loading.css` is produced and that `vsce ls` includes it in the VSIX), and lint, types and tests all pass.

I'm requesting changes rather than merging because the stated goal is to establish the pattern the other webviews will copy, and three things in the pattern itself need fixing first.

## Before merge

**1. Drop `'unsafe-inline'` from `style-src`** — `src/webview/webviewHtml.ts:56`

This is the one that matters. The PR does the work of moving the CSS into an external file and then gives the benefit straight back. After this change the loading document has no `<style>` block and no `style=` attributes, so it needs exactly zero inline style capability.

The comment on line 48 says styles are allowed inline because "VS Code theming relies on it". That isn't so. The official webview guide ships `style-src ${webview.cspSource};` with no `'unsafe-inline'`, and states that the CSP "implicitly disables inline scripts and styles" and that extracting them is the best practice. Theme colours arrive as CSS custom properties (`var(--vscode-editor-foreground)`), which external stylesheets read fine — which is exactly what `loading.css` already does.

Leaving it in means every future webview built on this helper inherits a `style-src` that still permits injected `<style>` and `style=`, and the next person to inline a style never gets the violation that would have told them to move it out.

**2. `webviewHtml.ts` imports `vscode` at module scope** — `src/webview/webviewHtml.ts:1`

Its two siblings, `inlineMarkdown.ts` and `scriptData.ts`, both carry the "`vscode`-free and pure so the unit suite can drive it directly" header and both have unit tests. This new module breaks that. Adding `tests/unit/webviewHtml.test.ts` would pull `import * as vscode from 'vscode'` into the mocha runner, which can't resolve it outside the extension host — that aborts the entire run, not just the one file.

So the shared helper every future webview depends on is currently untestable. `createNonce` needs no `vscode` at all, and `cspMetaTag` only reads `webview.cspSource`. Taking that as a string (`cspMetaTag(cspSource: string, nonce: string)`) leaves `assetUri` as the only function needing `vscode`, and makes the other two unit-testable.

**3. The webview build hardcodes a single entry point** — `esbuild.js:57`

`entryPoints: ['src/webview/styles/loading.css']`, while the comment below it and the PR both present this as a reusable pattern. Add `src/webview/styles/details.css` tomorrow and stylelint lints it, `tsc` passes, `node esbuild.js` exits 0, CI is green, and the file is simply never emitted — the `<link>` 404s silently in the packaged extension. A glob would fail loudly instead. Worth noting too that `outdir` is never cleaned, so a renamed stylesheet leaves a stale copy behind that still resolves.

## Scope

The CSP currently lands on 1 of the 6 webview documents in `componentBrowserProvider.ts`, and it's the loading spinner: no scripts, no remote data. The other five still emit inline `<script>` and interpolate component names, descriptions and tag lists fetched from GitLab — including the client-side markdown renderer. Those are the documents a CSP would actually protect.

That's fine as a first step, but the PR reads as though the webview is now under CSP. Either extend it to the remaining five, or reword so it's explicitly step one with a follow-up issue.

## Smaller things

- `esbuild.js:67` — the same `esbuildProblemMatcherPlugin` instance is registered on both contexts, so every build prints its markers twice. I ran it: two `[watch] build started`, two `[watch] build finished`. Since `Promise.all` interleaves them, a background problem matcher keyed on the first end marker will call the task ready while the second build is still going.
- `componentBrowserProvider.ts:92` — the `media` resource root is dead. No `media/` directory exists anywhere in the repo (icons are in `images/`), and this is its only reference. The diff already rewrites this array; good moment to drop it.
- `package.json:285` — `lint:ci` hardcodes a path into a dependency's internals (`node_modules/@csstools/stylelint-formatter-github/index.mjs`). That's correct for npm's flat layout only; it breaks under pnpm, under a nested version conflict, and if the script runs from a subdirectory. Passing the package name lets normal resolution handle it.
- `package.json:285` — the `&&` between eslint and stylelint means a PR with both a JS error and CSS errors only ever gets the JS annotations, then a second round trip for the CSS. Running both and aggregating exit codes gets the one-shot review this is aiming for. Relatedly, `lint` and `lint:ci` now duplicate the same two invocations, and CI no longer runs `lint` at all, so they will drift; `npm-run-all` is already a devDependency if you want to compose them.
- `componentBrowserProvider.ts:2500` — `getLoadingHtml` used to return a constant string, so the two back-to-back loading assignments in the cache-reset path were idempotent. Each call now embeds a fresh nonce, so the document is torn down and rebuilt twice and the spinner visibly restarts. Since this document has no scripts, hoisting the nonce or omitting `script-src` here restores a stable string.
- `src/webview/webviewHtml.ts:26` — `byte % 62` over bytes 0-255 is biased: `A`-`H` come up about 25% more often than the rest. Not exploitable at 32 characters, but the doc comment stakes a cryptographic claim, so either rejection-sample or use an alphabet whose size divides 256 (hex or base64url).
- `src/webview/webviewHtml.ts:40` — `assetUri` keeps `..` segments after `filter(Boolean)`, so `'../extension.js'` escapes the `out/webview` root the caller assumes, and `''` silently returns the directory. Unreachable from today's single literal call site, but the point of the module is that future callers pass computed paths.

## Docs

`AGENTS.md` designates `.ai/` as the specification, and this PR leaves it stale: `.ai/workflows.yaml:147` still says `lint: npm run lint` while CI now runs `lint:ci`, and `.ai/architecture.yaml:81` still documents a `src/templates/` directory that no longer exists while never mentioning `src/webview/` at all.

X-Guardian:
### Changes:

- CSP: dropped `unsafe-inline` from `style-src` in `/src/webview/webviewHtml.ts`. Confirmed the loading document has no `<style>` and no `style=`, and that `loading.css` reads theme colours via `var(--vscode-*)` — four occurrences — which works fine from an external file. Rewrote the comment, since my original claim that "VS Code theming relies on it" was wrong and would have taught the next person the same thing.
- Build: one marker pair, not two. The same plugin instance was on both contexts. Now the extension build emits the markers and the webview build uses a label-only error reporter. Verified: marker count went 2 → 1. I also guarded `location`, which is optional on esbuild errors — the original would have thrown on an error without one.
- Build: entry points discovered, not listed. I'd argued for an explicit list when planning; the reviewer's "silent-and-wrong beats loud-and-wrong" is the better argument. Worth noting my first attempt used esbuild's own glob syntax and broke the build — esbuild requires the directory to exist, and `src/webview/client/` doesn't yet. Resolving in Node instead handles missing directories. Verified both ways: adding a stylesheet picks it up with no config edit; removing one no longer leaves a stale copy, via an `rmSync` of `outdir`.
- Lint scripts. Split into `lint:eslint` / `lint:stylelint`, composed with `npm-run-all -c` (already a devDependency). Formatter now referenced by package name — verified it resolves through the package's exports map. Verified the aggregation: with both a JS and a CSS error present, both are reported in one run and the script exits non-zero. The old && showed only the JS one.
- Dead `media` root removed — single reference, no such directory.
- `.ai/ docs` — replaced the `src/templates/` block (directory gone) with an accurate `src/webview/` entry; noted `lint:ci` in workflows.

### Left for a follow-up, as proposed

The `webviewHtml.ts` `vscode` split plus unit tests, the nonce bias (`256 % 62 = 8`, so `A–H` are over-represented), the `assetUri` traversal guard, and the nonce hoisting that makes the spinner restart.

eFAILution (approved):
Approving. The two before-merge items that were about the pattern itself are properly resolved:

**1.** `'unsafe-inline'` is gone from `style-src`, and the comment that justified it is rewritten rather than just deleted — so the next person reads why a document that inlines a style is supposed to break.

**3.** Entry points are discovered rather than listed, and `outdir` is cleared each build, which closes the other half of that problem: a renamed stylesheet can no longer leave a stale copy behind that still resolves.

Smaller things all landed too — one marker pair with `location` guarded, the dead `media` root dropped, the stylelint formatter referenced by package name, both linters aggregated under `npm-run-all -c` so a JS error and a CSS error come back in one pass, and `.ai/` brought back in step.

Scope is reworded to "step one of six", which is what I was after. Nothing here claims more than it does.

**2** — the `vscode` split that makes `webviewHtml.ts` unit-testable — is the one left open, and taking it as a follow-up is reasonable while the module has a single literal call site. But the follow-up list lives in the PR body, and that stops being reachable the moment this squashes. Please open issues for the five, particularly the `vscode` split and the `assetUri` traversal guard: both get materially harder once the other five documents start calling into this helper with computed paths, which is the next PR in the sequence.

Verified before merging: base `beta`, mergeable and clean, `test (22.x)` and `extension-host` both green.

eFAILution:
Opened the follow-ups so they outlive this PR body:

- #284 — make `webviewHtml.ts` `vscode`-free and unit-testable (`cspMetaTag(cspSource, nonce)`)
- #285 — `createNonce` modulo bias
- #286 — `assetUri` `..` traversal guard
- #287 — per-call nonce restarts the spinner on cache reset
- #288 — extend the CSP to the remaining five documents

#288 is the sequence this PR is step one of; the other four are the helper changes its call sites depend on.

PR #277, opened 2026-09-08
fix(parser): add YAML merge key support to GITLAB_CI_SCHEMA (#268)

### LINKED ISSUE #268: fix(parser): GITLAB_CI_SCHEMA does not merge YAML merge keys (`<<:`)
Follow-up from #263, which introduced `GITLAB_CI_SCHEMA` in `src/utils/yamlParser.ts`.

## Problem

`GITLAB_CI_SCHEMA` is `yaml.CORE_SCHEMA.withTags(referenceTag)`, and neither `CORE_SCHEMA` nor js-yaml's own default carries the YAML merge tag. So `<<: *anchor` is parsed as a literal key named `<<` instead of merging the anchored mapping in.

GitLab parses CI YAML with Ruby's Psych (YAML 1.1), where `<<` **does** merge. Anchor + merge is a very common way to share config between jobs, so the extension's view of a pipeline diverges from GitLab's on real files.

This is **not a regression from #263** — js-yaml's default schema behaves the same way, so it predates that PR. But #263 is what established "the canonical GitLab CI schema", which makes this the natural place to fix it.

## Reproduction

Verified against `js-yaml@5.4.1` as pinned on `beta`:

```js
const text = 'a: &a\n  x: 1\nb:\n  <<: *a\n  y: 2\n';

GITLAB_CI_SCHEMA          => {"a":{"x":1},"b":{"<<":{"x":1},"y":2}}   // wrong
js-yaml default           => {"a":{"x":1},"b":{"<<":{"x":1},"y":2}}   // same, pre-existing
CORE_SCHEMA + mergeTag    => {"a":{"x":1},"b":{"x":1,"y":2}}          // correct
YAML11_SCHEMA             => {"a":{"x":1},"b":{"x":1,"true":2}}       // merges, but corrupts keys
```

## Downstream impact

`parseInputs` in `src/providers/localComponentResolver.ts` reads `definition.default === undefined` to decide whether an input is required. A `spec.inputs` entry that inherits its `default` through a merge key therefore gets reported as **required** when GitLab considers it optional, and loses its `type`, `description` and `options` at the same time. If `spec.inputs` itself is the merge target, completion offers an input literally named `<<`.

## Suggested fix

```ts
export const GITLAB_CI_SCHEMA = yaml.CORE_SCHEMA.withTags(yaml.mergeTag, referenceTag);
```

`js-yaml@5` exports `mergeTag`, so this is a one-line change.

**Please do not reach for `YAML11_SCHEMA` here.** It fixes merging but drags in the rest of YAML 1.1's scalar resolution: as the table above shows, the key `y` resolves to boolean `true`. `yes`, `no`, `on` and `off` go the same way, and all of those are plausible job or input names in a `.gitlab-ci.yml`.

Worth a unit test in `tests/unit/yamlParser.test.ts` covering both the merge and the fact that a `y`/`no`/`on` key survives as a string.

Cid-oe:
## Summary
Resolves #268.

GitLab CI uses Psych (YAML 1.1) which merges `<<: *anchor` keys into mappings. In `gitlab-component-helper`, `GITLAB_CI_SCHEMA` extended `yaml.CORE_SCHEMA` with only `referenceTag`, causing merge keys to parse as literal `<<` properties instead of merging anchored fields.

## Changes
- Updated `GITLAB_CI_SCHEMA` in `src/utils/yamlParser.ts` to include `yaml.mergeTag` alongside `referenceTag`.
- Retained `CORE_SCHEMA` base to ensure boolean-like strings (`y`, `no`, `on`, `off`) remain plain strings rather than resolving to booleans.
- Added unit tests in `tests/unit/yamlParser.test.ts` verifying:
  - Merge key (`<<:`) properly merges properties from anchored mappings.
  - Boolean-like keys (`y`, `no`, `on`) are preserved as string keys in the resulting mapping.

## Validation
- `npm test` passed cleanly (388/388 tests passing).

### DIFF
--- a/src/utils/yamlParser.ts
+++ b/src/utils/yamlParser.ts
@@ -20,8 +20,8 @@ const referenceTag = yaml.defineSequenceTag<unknown[]>('!reference', {
   identify: () => false,
 });
 
-/** The core schema plus GitLab's CI-only tags, so a `.gitlab-ci.yml` using them still parses structurally. */
-export const GITLAB_CI_SCHEMA = yaml.CORE_SCHEMA.withTags(referenceTag);
+/** The core schema plus GitLab's CI-only tags and YAML merge keys (`<<:`), so a `.gitlab-ci.yml` parses structurally. */
+export const GITLAB_CI_SCHEMA = yaml.CORE_SCHEMA.withTags(yaml.mergeTag, referenceTag);
 
 /** Type-guard: a parsed YAML value is a non-null object (i.e. a mapping). Use to narrow `unknown` results. */
 export function isYamlNode(value: unknown): value is YamlNode {
--- a/tests/unit/yamlParser.test.ts
+++ b/tests/unit/yamlParser.test.ts
@@ -66,6 +66,30 @@ test:
     const docs = parseYamlDocuments('test:\n  script:\n    - !reference [.setup, script]\n', true);
     assert.deepStrictEqual(docs[0].test, { script: [['.setup', 'script']] });
   });
+
+  test('merges YAML merge keys (<<:) and preserves y/no/on as string keys', () => {
+    const text = `
+a: &common
+  image: alpine
+  y: 10
+  no: false
+  on: true
+
+b:
+  <<: *common
+  stage: test
+`;
+    const docs = parseYamlDocuments(text, true);
+    assert.strictEqual(docs.length, 1);
+    assert.deepStrictEqual(docs[0].b, {
+      image: 'alpine',
+      y: 10,
+      no: false,
+      on: true,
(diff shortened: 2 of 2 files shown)


eFAILution:
Thanks @Cid-oe, and sorry for the collision. This is the correct fix, but #274 landed the same one-line change to `GITLAB_CI_SCHEMA` first, so I'm closing this as a duplicate rather than because anything is wrong with it.

For what it's worth you and #274 independently arrived at the same call, including covering the `y`/`no`/`on` key trap, which is the part that is easy to miss. Closing as superseded by #274.

PR #278, opened 2026-09-08
fix(providers): match cached component templatePath regardless of ref (#276)

### LINKED ISSUE #276: bug: component: template links don't appear for tag-pinned includes when the cache holds a different ref
### Bug summary

The document-link provider turns a `component:` include URL into a clickable link to the template file on GitLab. To build that link it looks the component up in the cache to get its `templatePath` (the repo layout). The lookup matched on `gitlabInstance` + `sourcePath` + `name` **and** version — requiring the cached entry's `version` to equal the version parsed from the include line.

That version clause is wrong for this purpose. `templatePath` is the repo layout, which is identical across refs; the produced link is already pinned separately to the include's `parsed.version`. So when the cache holds a component under one ref (say `main`, or the latest tag from discovery) but the include pins a different tag (`markdown-lint-1`), the lookup finds no match and **no link is produced** — even though the extension has everything it needs to build one.

Net effect: Ctrl/Cmd-click to "open template" silently fails for version-pinned includes whenever the cached ref differs from the pinned one.

### Steps to reproduce

1. Configure a component source under `gitlabComponentHelper.componentSources` so its components get cached by catalog discovery (which caches under the discovered/latest ref).
2. In a `.gitlab-ci.yml`, add an include pinned to a *specific tag* that isn't the cached ref:

   ```yaml
   include:
     - component: $CI_SERVER_FQDN/my-group/my-components/markdown-lint@markdown-lint-1
   ```

3. Hover / Ctrl+click the URL. Observe there is no document link (no underline, no "Open … template" tooltip, nothing to click).
4. Change the pin to match the cached ref exactly; the link now appears — confirming the version clause, not the component identity, was the gate.

### Expected behavior

The template link should appear for any include whose component is in the cache, regardless of which ref the cache entry was stored under. The link's ref comes from the include line (`parsed.version`), not from the cache entry, so a version match on the cache lookup is unnecessary.

### Actual behavior

- [documentLinkProvider.ts](src/providers/documentLinkProvider.ts) gated the cache lookup with `(!parsed.version || c.version === parsed.version)`. When `parsed.version` was set and no cached entry shared that exact ref, `.find` returned `undefined`, the `!cached?.templatePath` guard hit `continue`, and no link was pushed for that line.
- The link target is built by `templateFileUrlForResolved({ …, version: parsed.version, templatePath: cached.templatePath })` — the ref is taken from the include, so a differently-versioned cache entry would still have produced a correct link had the lookup matched.

Affects tag-pinned `component:` includes where the pinned ref differs from the ref the component was cached under — common in tag-per-component monorepos and any project where discovery caches the latest tag but consumers pin an older one.

```yaml
include:
  - component: $CI_SERVER_FQDN/my-group/my-components/markdown-lint@markdown-lint-1
```

### Extension logs

```text
N/A
```

### Affected area

Component Browser

### Extension version

0.15.5

### VS Code version

1.136.1

### Node.js version

24.0.0

### Operating system

macOS

### Pre-submit checks

Cid-oe:
## Summary
Resolves #276.

Allows document links for `component:` includes to resolve template paths from the component cache even when the include's pinned ref differs from the ref the component was discovered/cached under.

## Changes
- Removed the strict version matching filter (`(!parsed.version || c.version === parsed.version)`) when looking up cached components for `templatePath` resolution in `src/providers/documentLinkProvider.ts`.
- The resolved document link URL uses `parsed.version` from the include line directly (`templateFileUrlForResolved({ ..., version: parsed.version, templatePath: cached.templatePath })`), ensuring that repo layout (`templatePath`) is reused across refs without requiring an exact version match in cache discovery.

## Validation
- `npm test` passed (387 passing).
- `npm run compile` passed cleanly.

### DIFF
--- a/src/providers/documentLinkProvider.ts
+++ b/src/providers/documentLinkProvider.ts
@@ -88,8 +88,7 @@ export class ComponentDocumentLinkProvider implements vscode.DocumentLinkProvide
         (c) =>
           c.gitlabInstance === parsed.gitlabInstance &&
           c.sourcePath === parsed.path &&
-          c.name === parsed.name &&
-          (!parsed.version || c.version === parsed.version)
+          c.name === parsed.name
       );
 
       // Only produce a link when the cache has a resolved templatePath for the component. Uncached components are


eFAILution (approved):
Confirmed the reasoning holds. The link's ref comes from `parsed.version` off the include line, and `templateFileUrlForResolved` only takes `templatePath` from the cache entry — that's repo layout, identical across refs — so dropping the version clause turns "no link" into a correctly-pinned link rather than a wrong one.

Two notes, neither blocking:

- If a template file is renamed between refs, the borrowed `templatePath` could 404 at the pinned ref. Still strictly better than today, where you get nothing at all.
- No test, and there's no existing harness for `documentLinkProvider` — no unit or extension-host coverage — so adding one means a new extension-host suite. Worth a follow-up rather than holding this up.

Thanks @Cid-oe.

PR #280, opened 2026-09-09
fix(details): make Refresh Versions work in the browser-opened details panel

### LINKED ISSUE #279: bug: component browser: Refresh Versions silently does nothing in the details panel
### Bug summary

In the component **details** panel opened from the Component Browser's **Details** button, the **Refresh Versions** button does nothing. The version dropdown is unchanged, and no error is shown anywhere.

The panel's `fetchVersions` handler guards on a type-guard that requires the component to carry `url`. The component the panel receives is built client-side from `window.componentVersionData[name][version]`, a `ComponentVersion` — a type that has no `url` field. The guard therefore always fails on this path, the handler throws, and it replies `versionsError`. The webview's handler for that message only hides the spinner; its body is literally `// Could show error message here`. So the failure is total and completely silent.

`url` is required by the guard but **never read** by the code it gates: `fetchComponentVersions` matches the cache on `name` + `sourcePath` + `gitlabInstance`, and `versionCache.fetchComponentVersions` calls `fetchProjectTags(gitlabInstance, sourcePath)` / `fetchProjectDefaultBranch(gitlabInstance, sourcePath)`. Nothing in the version-fetch path touches `url` or `source`.

### Steps to reproduce

1. Configure a component source under `gitlabComponentHelper.componentSources` and let discovery populate the cache.
2. Run **GitLab CI: Browse Components**.
3. Click **Details** on any component.
4. In the details panel, click **Refresh Versions**.
5. Observe: nothing happens. The dropdown is unchanged and no error appears anywhere in the UI. (The `Loading version details...` label is shown and hidden again too quickly to see.)

The detached panel (hover → **View Full Component Details**) is *not* affected in the same way, because a hover-detected component carries the full cached shape including `url`.

### Expected behavior

**Refresh Versions** re-fetches the component's tags and repopulates the dropdown. Failing that, a failure should be visible — the button should not appear to work while doing nothing.

### Actual behavior

- `viewDetailsById` (in the browser webview script) builds the payload as `{ ...componentVersionData[name][version], name, version }`. `ComponentVersion` carries `version`, `description`, `parameters`, `documentationUrl`, `source`, `sourcePath`, `gitlabInstance` — **no `url`**.
- `showComponentDetails` enriches via `lookupComponentDetails`, which returns only `availableVersions` and `tagPattern`. `url` is never restored.
- The `fetchVersions` handler calls `isCachedComponentShape(component)`, which requires `source`, `sourcePath`, `gitlabInstance`, `version` **and `url`** to all be strings. Four of the five are present; `url` is not, so the guard returns `false`.
- The handler throws `Component is missing required fields (source, sourcePath, gitlabInstance, version) for version lookup.` and posts `versionsError`. Note the message does not even mention `url`, so the log is misleading if anyone gets that far.
- The webview's `versionsError` case hides the spinner and re-enables the select. No message is surfaced to the user.

## Suggested fix

Two candidates, and the choice is really "is `url` needed for a version lookup?":

1. **Relax the guard** for this call. Nothing in the version-fetch path reads `url` or `source`, so requiring them here is over-strict. A narrower guard (`sourcePath` + `gitlabInstance` + `version`, matching what `fetchComponentVersions` actually uses) would let the browser path work and keep the type honest about its real requirements.
2. **Populate `url`** on the way to the panel — either add it to `ComponentVersion`, or have `lookupComponentDetails` restore it from the matched cache entry alongside `availableVersions`/`tagPattern`.

(1) looks correct rather than merely expedient: the guard currently describes a shape the callee does not need, which is what makes it reject a component it could have served.

Independently, the webview's `versionsError` branch should surface something. A silent no-op is worse than an error message, and the `// Could show error message here` comment suggests this was known and deferred. Note this is what makes the bug hard to notice *or* report: because the failure is synchronous, the `Loading version details...` label never stays up long enough to register, so the button reads as inert rather than broken.

Affects the details panel opened from the Component Browser for any source. Most visible on tag-per-component monorepos, where the dropdown is the main way to move between versions.

```yaml
N/A
```

### Extension logs

**None — and that is part of the bug.** The `fetchVersions` handler's `catch` posts a `versionsError` message to the webview and does nothing else; there is no `logger` call on that path, at any log level. Setting `gitlabComponentHelper.logLevel` to `DEBUG` produces no entry for the failed refresh.

The thrown `Error` carries the message `Component is missing required fields (source, sourcePath, gitlabInstance, version) for version lookup.`, but it only ever reaches the webview's `versionsError` branch, which discards it. So the string exists in the source and is never surfaced anywhere a user or maintainer can see it — not the UI, not the output channel. (It would also be misleading if it were logged: it lists the four fields that *are* present and omits `url`, the one that is missing.)

### Affected area

Component Browser

### Extension version

0.17.3

### VS Code version

1.136.1

### Node.js version

24.0.0

### Operating system

macOS

### Pre-submit checks

X-Guardian:
## Summary
- **Refresh Versions** in the component details panel opened from the Component Browser did nothing at all — no dropdown change, no error in the UI, and no entry in the output channel. It failed on a type-guard that required a field the lookup never reads.
- The guard demanded `url`. The details panel's component is rebuilt in the webview from a `ComponentVersion`, which has no `url` field, so the guard rejected every component on that path before any network call.
- Replaces it with a guard describing what a version lookup actually needs, and makes the failure visible: the panel now reports the error, and the handler logs it.
- Link related issue(s): Fixes #279

## Change Type
- [ ] feat
- [x] fix
- [ ] refactor
- [ ] docs
- [ ] test
- [ ] chore

### User-facing impact
- **Refresh Versions works** in the details panel opened from the Component Browser. Previously it was inert for every component on that path.
- **A failed version refresh is now visible.** The panel shows `Could not load versions: …` next to the dropdown in the theme's error colour, and the reason is written to the output channel. Previously both were silent.
- The detached (hover) panel was already working, since a hover-detected component carries the full cache shape including `url`. Its error reporting improves the same way.

### Why the guard was wrong, not just strict
Nothing in the version-fetch path reads `url` or `source`:

- `ComponentCacheManager.fetchComponentVersions` matches the cache on `name` + `sourcePath` + `gitlabInstance`.
- `VersionCache.fetchComponentVersions` calls `fetchProjectTags(gitlabInstance, sourcePath)` and `fetchProjectDefaultBranch(gitlabInstance, sourcePath)`, then scopes the result with `tagPattern`.

So the guard described a shape the callee does not need, which is what made it reject components the cache could have served. The fix narrows the requirement to the fields actually used rather than loosening a check that was doing real work.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (no change to what is requested from GitLab; only which components are allowed to request it)

### Affected areas
- [x] Component Browser (details panel)
- [x] Hover provider (detached details panel — error reporting only)
- [ ] Completion provider
- [ ] Validation provider
- [x] Cache and refresh behavior (version lookup entry conditions)
- [ ] Docs only

## What changed by bucket

| Bucket | Files | Approach |
|---|---|---|
| The requirement, named | `src/types/cache.ts` | New `VersionLookupComponent` — `Pick<CachedComponent, 'name' \| 'sourcePath' \| 'gitlabInstance' \| 'version' \| 'tagPattern'>`. Names the subset a lookup needs instead of leaving callers to infer it from a full cache entry. |
| Signatures | `src/services/cache/componentCacheManager.ts`, `src/services/cache/versionCache.ts` | Both `fetchComponentVersions` take `VersionLookupComponent` rather than `CachedComponent`. No body changes — they already only read those fields. |
| The guard | `src/services/component/versionLookupShape.ts` | New `isVersionLookupShape`, checking `sourcePath` + `gitlabInstance` + `version`. Replaces `isCachedComponentShape`, which was **duplicated** in `src/providers/componentBrowserProvider.ts` and `src/extension.ts` and required `url` + `source` on top. Extracted to one `vscode`-free module so both callers share it and it is unit-testable. |
| Call sites | `src/providers/componentBrowserProvider.ts`, `src/extension.ts` | Both use the shared guard. The thrown message changes from a list of the four fields that *were* present to one naming what is actually missing. |
| Visible failure | `src/providers/componentBrowserProvider.ts` | The webview's `versionsError` branch — previously three lines ending in `// Could show error message here` — now shows the reason in the existing status slot, in `--vscode-errorForeground`. Cleared when a refresh starts and when one succeeds, so a stale error cannot persist. |
| Logged failure | `src/providers/componentBrowserProvider.ts`, `src/extension.ts` | Both `fetchVersions` catch blocks now log before posting `versionsError`. Neither logged anything before, so the failure left no trace at any log level. |
| Unit coverage | `tests/unit/versionLookupShape.test.ts` | Six cases. The load-bearing one asserts a component with **no `url`** is accepted — that is the exact shape the details panel receives, and the assertion fails against the old guard. |

### Local checks
- [x] `npm run compile`
- [x] `npm test` — 402 passing (+6 new); `npm run lint` clean
- [x] Extension-host suite — 25 passing locally

### Manual test notes
- To verify the fix:
  - **GitLab CI: Browse Components** → **Details** on any component → **Refresh Versions** → the dropdown repopulates. *(Does nothing on `beta`.)*
  - Best seen on a tag-per-component monorepo source (one with a `tagPattern`), where the dropdown is the main way to move between versions.
- To verify the error path, point a source at an unreachable instance or revoke its token, then click **Refresh Versions**: the panel should show `Could not load versions: …` and the output channel should carry a matching `[ComponentBrowser] Error fetching versions:` line. *(Both silent on `beta`.)*
- Confirm the error clears: trigger a failure, then fix the cause and refresh again — the message should be replaced by the loading text, not left behind.

## Breaking Changes
- [x] No breaking changes

## Risk and Rollback
- Main risks: **low**. The change widens what is accepted into an existing read-only lookup; it adds no new call, request or write. A component that passed the old guard still passes the new one — `isCachedComponentShape` required a strict superset of these fields — so no path that worked before can start failing.
- The one behavioural consequence is intended: components that were previously rejected now reach `fetchComponentVersions`. That call already handles its own failure, returning `[component.version]` as a fallback and logging on error.
- Rollback strategy: revert the commit. No persisted state, settings, cache shape or webview protocol changes; `versionsLoaded`/`versionsError` keep their existing field names.

## Release Notes Draft
- Fix: **Refresh Versions** now works in the component details panel opened from the Component Browser; it previously did nothing.
- Fix: a failed version refresh now reports the reason in the panel and the output channel instead of failing silently.

## Checklist
- [ ] Branch is up to date with target branch
- [ ] Commit messages follow conventional commits
- [x] Added/updated docs for behavior or settings changes (this PR description; no user-facing settings changed)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -10,28 +10,13 @@ import { getComponentCacheManager, ComponentCacheManager } from './services/cach
 import { Logger } from './utils/logger';
 import { ValidationProvider } from './providers/validationProvider';
 import type { CachedComponent } from './types/cache';
+import { isVersionLookupShape } from './services/component/versionLookupShape';
 import type { GitLabYamlFragment } from './types/gitlab-catalog';
 import type { HoverContext } from './providers/hoverContentBuilder';
 
 /** Component payload passed to the `detachHover` command. Adds the hover-builder's location context. */
 type DetachableComponent = Component & { _hoverContext?: HoverContext };
 
-/**
- * Type-guard narrowing a `Component`-shaped value to one that also satisfies `CachedComponent`.
- *
- * @param component  A `Component` (typically `activeComponent` in the detach-hover panel) that may
- *                   or may not have been enriched with cache details.
- * @returns          `true` if all `CachedComponent` required fields are present and string-typed,
- *                   narrowing `component` to `Component & CachedComponent` in the truthy branch.
- *                   `false` if any field is missing.
- */
-function isCachedComponentShape(component: Component): component is Component & CachedComponent {
-  return typeof component.source === 'string'
-    && typeof component.sourcePath === 'string'
-    && typeof component.gitlabInstance === 'string'
-    && typeof component.version === 'string'
-    && typeof component.url === 'string';
-}
 import { getPerformanceMonitor } from './utils/performanceMonitor';
 import { isGitLabCIFile, invalidateFileGlobsCache } from './utils/gitlabCiFileMatcher';
 
@@ -509,29 +494,33 @@ export function activate(context: vscode.ExtensionContext) {
               break;
             case 'fetchVersions':
               try {
-                if (!isCachedComponentShape(activeComponent)) {
-                  throw new Error('Component is missing required fields (source, sourcePath, gitlabInstance, version) for version lookup.');
+                if (!isVersionLookupShape(activeComponent)) {
+                  throw new Error(`Cannot look up versions for ${activeComponent.name}: missing source path or GitLab instance.`);
                 }
-                const versions = await cacheManager.fetchComponentVersions(activeComponent);
+                // Bind the narrowed value: `activeComponent` is reassignable, so TS widens it back across the await.
+                const lookupTarget = activeComponent;
+                const versions = await cacheManager.fetchComponentVersions(lookupTarget);
                 // For a monorepo source, map each full tag to its stripped {version} so the dropdown shows short
                 // labels while keeping the full tag as the option value (the inserted ref).
                 let versionLabels: Record<string, string> | undefined;
-                if (activeComponent.tagPattern) {
+                if (lookupTarget.tagPattern) {
                   versionLabels = {};
                   for (const v of versions) {
-                    versionLabels[v] = stripTagPrefix(v, activeComponent.name, activeComponent.tagPattern);
+                    versionLabels[v] = stripTagPrefix(v, lookupTarget.name, lookupTarget.tagPattern);
                   }
                 }
                 panel.webview.postMessage({
                   command: 'versionsLoaded',
                   versions: versions,
                   versionLabels,
-                  currentVersion: activeComponent.version
+                  currentVersion: lookupTarget.version
                 });
               } catch (error) {
+                const reason = error instanceof Error ? error.message : String(error);
+                logger.error(`[Extension] Error fetching versions for detached details panel: ${reason}`, 'Extension');
                 panel.webview.postMessage({
                   command: 'versionsError',
-                  error: error instanceof Error ? error.message : String(error)
+                  error: reason
                 });
               }
               break;
--- a/src/providers/componentBrowserProvider.ts
+++ b/src/providers/componentBrowserProvider.ts
@@ -3,7 +3,7 @@ import { getComponentService } from '../services/component';
 import { ComponentCacheManager } from '../services/cache/componentCacheManager';
 import { GitLabCatalogComponent, GitLabCatalogVariable } from '../types/gitlab-catalog';
 import type { ComponentParameter, Component } from './componentDetector';
-import type { CachedComponent } from '../types/cache';
+import { isVersionLookupShape } from '../services/component/versionLookupShape';
 import type { SourceGroup, ComponentGroup, ComponentVersion } from './componentBrowserTypes';
 import type { HoverContext } from './hoverContentBuilder';
 import { containsGitLabVariables } from '../utils/gitlabVariables';
@@ -24,19 +24,6 @@ import { compileTagTemplate, stripTagPrefix } from '../services/component/tagSco
  */
 type DetachableComponent = Component & { _hoverContext?: HoverContext };
 
-/**
- * Type-guard narrowing a `Component`-shaped value to one that also satisfies `CachedComponent`.
- * `Component` carries `source`/`sourcePath`/`gitlabInstance`/`version`/`url` as optional; cache
- * methods like `fetchComponentVersions` require them. The guard checks all five before the call so
- * we don't pass a half-populated `Component` into a function expecting the full cache shape.
- */
-function isCachedComponentShape(component: Component): component is Component & CachedComponent {
-  return typeof component.source === 'string'
-    && typeof component.sourcePath === 'string'
-    && typeof component.gitlabInstance === 'string'
-    && typeof component.version === 'string'
-    && typeof component.url === 'string';
-}
 
 /**
  * Pre-existing component shape parsed out of a `.gitlab-ci.yml` include line by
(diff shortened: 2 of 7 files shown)


eFAILution:
@X-Guardian heads up — I pushed one commit to this branch (9029845) rather than asking you for another round trip. Revert it if you disagree with any of it.

#281 was opened against the same issue and caught two things this branch missed, so I ported them here instead of merging a duplicate:

- **Monorepo labels.** The browser details panel's `fetchVersions` never sent `versionLabels`. The webview already knows how to use them, so on a tag-per-component source a refresh silently reverted the dropdown from `1.0.0` back to `deploy-1.0.0`. The detached panel had this right; the browser one didn't.
- **`versionChangeError`.** It still hid the spinner and said nothing — the same silent failure this PR exists to fix, one case block over. It now reports in the same slot as `versionsError`, and both entry points clear a stale error before retrying.

Rather than copy the label loop a third time, I moved it to `buildVersionLabels` in `tagScoping.ts` and collapsed the two existing copies onto it — same pure-and-unit-tested shape as the `versionLookupShape` module you added here. Four tests, and it compiles the tag template once per list instead of once per tag.

Everything green: lint, both tsconfigs, 406 unit tests, extension-host suite.

The rest of the PR I'd merge as-is. Deleting both copies of `isCachedComponentShape` and narrowing `fetchComponentVersions` to `VersionLookupComponent` is the right call — it fixes the type rather than the symptom, which is why I took this over #281.

eFAILution (approved):
Taking this over #281, which fixed the same issue. This one treats it as a type problem rather than patching the symptom: naming `VersionLookupComponent`, narrowing `fetchComponentVersions` to it, and deleting both copy-pasted `isCachedComponentShape` guards in favour of one shared, vscode-free, unit-tested module.

That matters beyond this bug. The old guard demanded `url` and `source`, which nothing on the lookup path reads, so it rejected components the cache could have served. The signature now states what the function actually needs.

Also picked up here: the missing `logger` call the issue called out, and a visible message where the refresh previously failed in total silence.

I pushed 9029845 on top with the two things #281 caught — monorepo `versionLabels` on the browser path, and the still-silent `versionChangeError` branch — plus `buildVersionLabels` to keep the label loop from existing in three places.

Issue #289, opened 2026-09-11
bug: component browser: changing a component version breaks the panel


X-Guardian:
### Bug summary

Changing a component's version leaves the webview showing the previous version's content, in both the details panel and the Component Browser list. Nothing is reported in the UI and the output channel stays silent, so the panel simply appears not to respond.

Affects every component on every source.

### Steps to reproduce

1. Configure any component source under `gitlabComponentHelper.componentSources`.
2. Run **GitLab CI: Browse Components**.
3. Click **Details** on any component.
4. Change the **Version** dropdown.

For the Component Browser list, stop at step 2 and change a component's version dropdown there instead.

### Expected behavior

Changing the version replaces the panel's description, parameters and Template File link with the selected version's, and clears the loading indicator.

### Actual behavior

In the details panel:

- `Loading version details...` appears beside the dropdown and never clears.
- The description, parameters and Template File link keep the previously selected version's values.
- The dropdown itself shows the newly selected version, so the panel presents one version's label against another version's content.

In the Component Browser list:

- The description is not updated.
- The "Selected version" line is not updated.
- **Insert** and **Details** continue to act on the previously selected version, so inserting after a version change writes the wrong ref into the file.

Both panels render correctly when first opened; only updates fail. Reopening the panel restores a consistent view.

### Component context (if applicable)

Independent of `tagPattern` and monorepo layout — both the tag-per-component and ordinary single-component sources behave the same way.

### Minimal .gitlab-ci.yml snippet

```yaml
N/A
```

### Extension logs

```text
The only record is in the webview's own devtools console (**Developer: Open Webview Developer Tools**, with the panel focused):


Version changed to: cloud-deploy-aws-ecs-2.0.0
Received message: {command: 'componentDetailsUpdated', component: {…}}
Updating component details: {name: 'cloud-deploy-aws-ecs', …}
Uncaught SyntaxError: Invalid regular expression: /[([^]]+)]((https?://[^s)]+))/g: Unmatched ')'
    at new RegExp (<anonymous>)
    at renderInlineMarkdown (<anonymous>:16:20)
    at updateComponentDetails (<anonymous>:124:73)
```

### Affected area

Component Browser

### Extension version

0.17.5

### VS Code version

1.136.1

### Node.js version

24.x

### Operating system

macOS

### Pre-submit checks

- [x] I searched existing issues and did not find a duplicate.
- [x] I can reproduce this on the latest version from main/beta.
[...]

PR #290, opened 2026-09-11
fix(webview): restore version switching in the component browser and details panel

X-Guardian:
## Summary
- Changing a component's version left the webview half-updated: the details panel showed `Loading version details...` indefinitely while keeping the previous version's description, parameters and Template File link, and the Component Browser list's version dropdown did nothing at all.
- The client-side copy of `renderInlineMarkdown` built its patterns with `new RegExp('…')`, which unescapes twice — once when the extension evaluates the template literal producing the source, once when the browser parses the resulting string literal. `\[` arrived as a bare `[`, and the pattern threw `Invalid regular expression` on construction, so the function failed on its first call for every component.
- Switches the patterns to regex literals, extracts the source into `src/webview/clientInlineMarkdown.ts`, and adds tests that evaluate the emitted source.
- Link related issue(s): Fixes #289

## Change Type
- [ ] feat
- [x] fix
- [ ] refactor
- [ ] docs
- [ ] test
- [ ] chore

### User-facing impact
- **Changing a version in the details panel works.** The description, parameters and Template File link update, and the loading indicator clears.
- **Changing a version in the Component Browser list works.** The description and "Selected version" line update, and the **Insert** and **Details** buttons are rewired to the selected version. That rewiring sat after the throw, so **Insert** previously wrote the previously selected ref into the file — the most damaging symptom, since it silently produces a wrong `.gitlab-ci.yml` rather than just a stale view.
- Affects every component on every source, regardless of `tagPattern` or description content. Initial renders were always correct, since they use the server-side module; only updates were broken.

### Why it threw for every component
`new RegExp` validates its pattern when the object is **constructed**, not when it matches. The invalid pattern was built on the first line of the return expression, so the function threw before examining any text — a plain-text description failed exactly like one containing a markdown link.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (webview rendering only; no GitLab interaction)

### Affected areas
- [x] Component Browser (list version dropdown, and the details panel from both entry points)
- [x] Hover provider (the detached "View Full Component Details" panel renders the same document)
- [ ] Completion provider
- [ ] Validation provider
- [ ] Docs only

## What changed by bucket

| Bucket | Files | Approach |
|---|---|---|
| The fix | `src/webview/clientInlineMarkdown.ts` | Patterns are regex literals, so the source is unescaped once instead of twice and mirrors `src/webview/inlineMarkdown.ts` line for line. |
| Extraction | `src/webview/clientInlineMarkdown.ts`, `src/providers/componentBrowserProvider.ts` | The source-text builder moves out of the provider into a `vscode`-free module, which is what makes it reachable from the unit suite — `componentBrowserProvider.ts` imports `vscode` and cannot be loaded outside the extension host. The provider method delegates, so both call sites are untouched. |
| Test coverage | `tests/unit/clientInlineMarkdown.test.ts` | Five cases that `new Function(...)` the emitted source and exercise the resulting renderer, including one comparing its output against the server module across representative descriptions so the twins cannot drift. |

### On testing the emitted source
The fault lives between the TypeScript source and the browser's parse of it, so `tsc`, eslint and any assertion on the string itself all pass while the function is broken at runtime. Evaluating the emitted source is the only check that fails.

### Local checks
- [x] `npm run compile`
- [x] `npm test` — 411 passing (+5 new); `npm run lint` clean
- [ ] Manual verification in VS Code Extension Host — **outstanding**

### Manual test notes
- Confirmed the tests are load-bearing: restoring `new RegExp('…')` fails all five with `Invalid regular expression: /[([^]]+)]((https?://[^s)]+))/g: Unmatched ')'` — the exact error reported from the webview console.
- Verified against the built bundle rather than the source: evaluated `clientRenderInlineMarkdownSource` out of `out/extension.js` and checked the emitted pattern is `/\[([^\]]+)\]\((https?:\/\/[^\s)]+)\)/g`, then exercised the function on plain, linked, code-span and HTML-bearing descriptions.

## Breaking Changes
- [x] No breaking changes

## Risk and Rollback
- Main risks: **low**. The change is confined to the pattern syntax inside one generated function, and the new tests pin its output to the server module's across the formatting and escaping cases. The escaping behaviour is unchanged — the same five HTML-significant characters, the same four formatting patterns.
- The code currently throws on every client-side call, so there is no working behaviour to regress.
- Rollback strategy: revert the commit. No persisted state, settings, cache shape or webview protocol changes.

## Release Notes Draft
- Fix: changing a component's version now updates the details panel and the Component Browser list. Both previously failed silently, leaving the panel showing the previous version with a stuck loading indicator.

## Checklist
- [x] Added/updated docs for behavior or settings changes (this PR description; no user-facing settings changed)


eFAILution (approved):
Approving. The diagnosis is exactly right and the fix is the minimal correct one.

## Verified the root cause

Reproduced both forms standalone, evaluating the emitted source the way the webview does:

```
OLD emitted: .replace(new RegExp('\[([^\]]+)\]\((https?://[^\s)]+)\)', 'g'), ...)
OLD result:  THREW: Invalid regular expression: /[([^]]+)]((https?://[^s)]+))/g: Unmatched ')'

NEW emitted: .replace(/\[([^\]]+)\]\((https?:\/\/[^\s)]+)\)/g, ...)
NEW result:  "see <a href=\"https://example.com/x\">docs</a>"
```

Byte-identical to the error you reported from the webview console. The two-stage unescape reading is correct: the template literal consumes one layer, the browser's parse of the resulting *string literal* consumes the second — and that second one is precisely what the old doc comment claimed `new RegExp` was avoiding. It was causing it.

Worth saying plainly, since the old comment will outlive this PR in `git log`: interpolating a runtime string into the HTML template literal performs no escape processing at all. There was nothing to defend against, which is why the defence broke it.

## Verified the impact

Read both client call sites rather than taking the description on trust:

- `componentBrowserProvider.ts:1265` — the throw lands before the version-info line, the Insert rewiring and the Details rewiring. Your read is right that the Insert button is the damaging one: it keeps the previously selected version's `onclick`, so the user picks a version, presses Insert, and gets a different ref in the file with no indication anything went wrong. That deserved the top billing you gave it.
- `componentBrowserProvider.ts:1935` — first statement of `updateComponentDetails` after the name, so nothing downstream runs, including whatever clears the loading indicator. Matches the reported symptom.

And the initial-render claim holds: `:741` and `:1720` both go through the server module, so only updates were ever broken.

## Verified the tests are load-bearing

Substituted the original line back in (taken verbatim from `20bef37`) and ran the suite: all five fail, with that same `Unmatched ')'`. Restored and re-ran: 411 passing.

The cross-check against the server renderer is the right guard here — the two implementations have no compiler relationship, and that test is what will catch the next drift.

## Also checked

- `grep -rn "new RegExp('" src/` — no other instance. This was the only place the pattern appeared.
- Placement: `src/webview/clientInlineMarkdown.ts` rather than `src/webview/client/`. Correct, and worth being deliberate about — `esbuild.js` treats `src/webview/client/*.ts` as browser entry points, and this module is Node-side code that *emits* browser source. It would have been bundled as its own asset in `client/`.
- Merges into current `beta` with no conflicts, so no rebase needed.

## Local checks

`npm test` 411 passing, `eslint --max-warnings 0` clean, `tsc --noEmit` clean. CI green, `extension-host` included.

## One note, not blocking

Manual Extension Host verification is still unchecked in the description. The `extension-host` suite doesn't drive webview scripts, so it isn't covering that gap. I'm merging anyway: the failure mode is a hard throw at `RegExp` construction rather than a subtle rendering difference, and the new tests evaluate the exact string that lands in the HTML. Worth a click through the version dropdown when you next have the host open, to confirm the panel clears.

Issue #299, opened 2026-09-15
bug: details panel: Template File link disappears after changing version in the detached view


X-Guardian:
### Bug summary

In the component details panel opened from a hover, changing the version removes the **Template File** row. The link does not come back until the panel is closed and reopened.

The same panel opened from the Component Browser's **Details** button keeps its link when the version changes.

### Steps to reproduce

1. Configure a component source under `gitlabComponentHelper.componentSources` and let discovery populate the cache.
2. In a `.gitlab-ci.yml`, hover a `component:` include and click **📄 View Full Component Details**.
3. Confirm the panel shows a **Template File:** row with a GitLab URL.
4. Change the **Version** dropdown.

### Expected behavior

The Template File row stays, its URL updated to point at the newly selected ref — as it does in the panel opened from the Component Browser.

### Actual behavior

The Template File row disappears entirely. Everything else updates: the description, parameters and Component URL all reflect the selected version.

Closing and reopening the panel restores the row.

### Component context (if applicable)

Needs a component whose cached entry carries a `templatePath`; without one the row is absent on the initial render too, so there is nothing to lose. Independent of `tagPattern` and monorepo layout.

### Minimal .gitlab-ci.yml snippet

```yaml
include:
  - component: $CI_SERVER_FQDN/my-group/my-components/my-component@1.0.0
```

### Extension logs

```text
Nothing is logged. The version change completes normally on the extension side; the row is hidden by the webview when it applies the reply.
```

### Affected area

Component Browser

### Extension version

0.17.10

### VS Code version

1.136.1

### Node.js version

24.x

### Operating system

macOS

### Pre-submit checks

- [x] I searched existing issues and did not find a duplicate.
- [x] I can reproduce this on the latest version from main/beta.
- [x] I included a minimal YAML snippet and relevant logs.

eFAILution:
Fixed by #300, merged to `beta` as 8f1a991. Reaches `main` with the next release.

The details panel had two copies of its message handler, one per entry point, and only the Component Browser's attached `templateFileUrl` to `componentDetailsUpdated` — so the detached (hover) panel dropped the link on a version switch, since the webview hides the element when the field is absent. Both entry points now share one handler.

Also closed a latent bug on the same path: the detached copy replaced its component on `versionChanged` without carrying `_hoverContext` forward, so switching version and then inserting would have added a duplicate include instead of editing the existing one in place.

PR #300, opened 2026-09-15
fix(details): share one message handler across both details-panel entry points

X-Guardian:
## Summary
- The component-details webview is opened two ways — the Component Browser's **Details** button and a hover's **View Full Component Details** — and each path carried its own full copy of the panel's message handler. Same three inbound commands, same four outbound, same HTML, but the copies had drifted and **each was missing a field the other sent**.
- Consolidates both into one `registerDetailsPanelMessageHandler` on `ComponentBrowserProvider`. Net −97 lines.
- Fixes the Template File link disappearing after a version switch in the detached panel, and closes a latent bug — a stale `currentVersion` reported after a version switch.
- Link related issue(s): #299

## Change Type
- [ ] feat
- [x] fix
- [x] refactor
- [ ] docs
- [ ] test
- [ ] chore

### User-facing impact
- **The Template File link no longer disappears.** In the detached (hover) panel, switching version dropped the link, because only the browser path attached `templateFileUrl` to `componentDetailsUpdated`. The webview hides the element when the field is absent, so the link vanished until the panel was reopened.
- **`Refresh Versions` after a version switch now reports the version on screen.** The browser path never replaced its component after `versionChanged`, so a later `fetchVersions` sent the originally-opened `currentVersion`.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (no change to GitLab interaction; message routing and payload shaping only)

### Affected areas
- [x] Component Browser (details panel, both entry points)
- [x] Hover provider (the detached "View Full Component Details" panel)
- [ ] Completion provider
- [ ] Validation provider
- [ ] Docs only

## What changed by bucket

| Bucket | Files | Approach |
|---|---|---|
| Shared handler | `src/providers/componentBrowserProvider.ts` | New `registerDetailsPanelMessageHandler(panel, component, options)` holding the merged behaviour of both copies, plus a `DetailsPanelOptions` interface. The component is held in a mutable local so a `versionChanged` round trip replaces it — the detached path's behaviour, now applied to both. `_hoverContext` is carried across that reassignment so an edit-in-place after a version switch still edits rather than inserting anew. |
| Browser call site | `src/providers/componentBrowserProvider.ts` | `showComponentDetails`'s inline `if/else-if` chain (~105 lines) collapses to one call. |
| Detached call site | `src/extension.ts` | The `switch` (~137 lines) collapses to one call passing `{ detachedFrom: originalEditor }`, dropping the imports it no longer needs. |
| Editor adoption | `src/providers/componentBrowserProvider.ts` | The detached path builds its own `ComponentBrowserProvider` and never calls `show()`, so `originalEditor` is unset and `insertComponent` would refuse to insert. The handler adopts the editor passed via `detachedFrom` at registration, so `insertComponent` behaves identically on both paths. |
| Dead code | `src/providers/componentBrowserProvider.ts` | `insertComponentFromDetached`, a one-line pass-through to `insertComponent`, has no callers left and is removed. |

### Local checks
- [x] `npm run compile`
- [x] `npm test` — 411 passing; `npm run lint` clean

### Manual test notes
- **Confirmed both ways.** Hover a `component:` include → **View Full Component Details** → switch version: the Template File link disappears on `beta` and is retained here. Needs a component with a cached `templatePath`, or the link is absent on initial render too.
- Insert on both paths, with and without inputs, checking what lands in the `.gitlab-ci.yml`: the browser path adds an include at the cursor, the detached path edits the existing one in place. Both now run the same handler, and nothing automated covers either write.

## Breaking Changes
- [x] No breaking changes

## Risk and Rollback
- Main risks: **medium**. One handler now picks between editing an include in place and adding a new one, keyed on `_hoverContext`. The wrong branch duplicates an include or overwrites the wrong lines, which only the insert checks above would catch.

## Release Notes Draft
- Fix: the Template File link no longer disappears after switching version in the detached component details view.

## Checklist
- [ ] Added/updated tests for new behavior (the handler imports `vscode` and is not reachable from the unit suite; the label logic it calls is covered by the existing `tagScoping` tests)


eFAILution (approved):
Approving. Two copies of a handler that had each drifted into missing something the other sent is exactly the shape worth collapsing, and the consolidation is faithful.

## Traced both paths rather than trusting the merge

The three behaviours I'd expect a consolidation like this to quietly change all hold:

- **Editor focus.** The detached copy did `showTextDocument(originalEditor…)` plus an activation wait and an active-editor check inline, before branching. That's gone from the call site, which reads like a loss — but `insertComponent` already does all three itself (`:366`–`:375`), and the edit branch goes to `editExistingComponentFromDetached`, which opens and focuses the hover's own document. The inline version was redundant on one branch and pointed at the wrong document on the other.
- **`adoptOriginalEditor` cannot clobber the browser's editor.** `extension.ts:326` constructs a fresh `ComponentBrowserProvider` per invocation of the detached command, so the adoption lands on an instance nothing else shares. Worth stating explicitly, because if the provider were a singleton this would silently retarget the browser's inserts.
- **Insert-versus-edit.** Keyed on `active._hoverContext` in both, as before. Browser-path components don't carry one, so that path is unchanged.

The two bug fixes are real and I'd have missed the second one: `templateFileUrl` was only ever attached on the browser path, and the old detached `versionChanged` did `activeComponent = updatedComponent`, dropping `_hoverContext` — so switching version and then inserting would have added a duplicate include instead of editing in place. That was latent and unreported. Good catch.

`insertComponentFromDetached` has no remaining callers. Checked.

411 passing, lint and types clean, CI green including extension-host.

## One thing I pushed a fix for

`DetailsPanelOptions.detachedFrom` documented itself as:

> Its presence means an insert refocuses that editor first and closes the panel afterwards; the Component Browser's own panel does neither.

The refocus half isn't right — both paths refocus, unconditionally, inside `insertComponent` and `editExistingComponentFromDetached`. What the option actually controls is adopting the editor as the insertion target at registration (because the detached entry point builds its own provider and nothing else populates it) and disposing the panel after an insert.

Worth correcting rather than leaving, because it's the doc the next caller of this handler reads to decide whether they need the option, and it describes a behaviour they'd get either way. `c465024`.

## Not blocking

No automated coverage, and your note on why is fair — the handler imports `vscode`. The part that's worth extracting one day is the insert-versus-edit decision, since that's the branch where being wrong duplicates or overwrites a user's include, and it's a pure function of `_hoverContext`. Not for this PR.

Issue #284, opened 2026-09-11
refactor(webview): webviewHtml.ts imports vscode at module scope, so the shared helper can't be unit-tested


eFAILution:
Follow-up from #275.

## Problem

`src/webview/webviewHtml.ts:1` imports `vscode` at module scope. Its two siblings, `src/webview/inlineMarkdown.ts` and `src/webview/scriptData.ts`, both carry a "`vscode`-free and pure so the unit suite can drive it directly" header and both have unit tests. This module breaks that, and it is the one every future webview document will depend on.

Adding `tests/unit/webviewHtml.test.ts` today would pull `import * as vscode from 'vscode'` into the mocha runner, which cannot resolve it outside the extension host. That aborts the entire run, not just the one file — the same failure mode `src/parsers/specParser.ts` already works around by importing from `errors/types` rather than the `errors` barrel.

So the shared helper the other five webview documents are meant to build on is currently the only one of the three that cannot be tested.

## Suggested fix

`createNonce` needs no `vscode` at all, and `cspMetaTag` only reads `webview.cspSource`. Taking that as a string:

```ts
export function cspMetaTag(cspSource: string, nonce: string): string
```

leaves `assetUri` as the only function that needs `vscode`, and makes the other two unit-testable. The single call site in `componentBrowserProvider.ts` passes `webview.cspSource` instead of `webview`.

Then add `tests/unit/webviewHtml.test.ts` covering the nonce length and alphabet, and the directive set the meta tag emits — including that `style-src` carries no `'unsafe-inline'`. That property is the whole point of #275 and nothing currently guards it against being reintroduced.

eFAILution:
Closed by #305, merged to `beta` as b236672. Reaches `main` with the next release.

Issue #288, opened 2026-09-11
refactor(webview): extend the CSP to the remaining five Component Browser documents


eFAILution:
Follow-up from #275.

## Problem

#275 put 1 of the 6 webview documents in `componentBrowserProvider.ts` under a Content-Security-Policy — deliberately the loading spinner, as the smallest and lowest-risk surface: no scripts, no remote data. The other five are the ones a CSP would actually protect:

- `getNoSourcesHtml` — `componentBrowserProvider.ts:311`
- `getErrorsHtml` — `componentBrowserProvider.ts:317`
- `getComponentBrowserHtml` — `componentBrowserProvider.ts:339`
- `getErrorHtml` — `componentBrowserProvider.ts:343`
- `getComponentDetailsHtml` — `componentBrowserProvider.ts:492`

They emit inline `<script>` blocks and interpolate component names, descriptions and tag lists fetched from GitLab, including through the client-side markdown renderer. #238 is open against exactly that surface — parameter descriptions and notes rendered unescaped — and a CSP is the layer that would contain it rather than relying on every interpolation site escaping correctly.

## Scope

For each document:

- extract the inline `<script>` into `src/webview/client/`. The esbuild config from #275 already discovers that directory and emits to `out/webview/client/`, so no build change is needed.
- convert inline event handlers to delegated listeners. A nonce CSP blocks `onclick=` and friends; there are 33 inline handlers across the five.
- emit the CSP meta tag and load the script via `assetUri` + the nonce, as `getLoadingHtml` now does for its stylesheet.

Likely one PR per document rather than one for all five, given the handler rewrite each needs.

## Depends on

The helper follow-ups from #275 should land first — #284 (unit-testable helper), #286 (`assetUri` guard) and #285 (nonce bias). These documents are the call sites that make them load-bearing.

eFAILution:
Progress: **4 of 6 documents are under a CSP.**

- `getLoadingHtml` — #275
- `getNoSourcesHtml`, `getErrorsHtml`, `getErrorHtml` — #308 (701c444)

Remaining, and the two that actually hold remote data:

- `getComponentBrowserHtml` (`componentBrowserProvider.ts:706`)
- `getComponentDetailsHtml`

[...]

Issue #286, opened 2026-09-11
fix(webview): assetUri keeps `..` segments, so a computed path can escape out/webview


eFAILution:
Follow-up from #275.

## Problem

`src/webview/webviewHtml.ts:40`:

```ts
const parts = relativePath.split('/').filter(Boolean);
return webview.asWebviewUri(
  vscode.Uri.joinPath(extensionUri, 'out', 'webview', ...parts)
);
```

`filter(Boolean)` drops empty segments but keeps `..`, so `assetUri(webview, extensionUri, '../extension.js')` resolves outside the `out/webview` root the doc comment promises. An empty `relativePath` silently returns the directory itself rather than failing.

Unreachable today — the only call site passes the literal `'styles/loading.css'`. It matters because the point of the module is that future callers pass computed paths, and the next PRs in the #275 sequence are exactly those callers.

## Suggested fix

Reject a path that contains a `..` segment, starts from the filesystem root, or yields no segments at all. Throw rather than return a URI: a bad asset path is a programming error, and a silently wrong URI just 404s inside the webview with nothing pointing back here.

Covered by the unit tests this module gains once it is `vscode`-free.

eFAILution:
Closed by #305, merged to `beta` as b236672. Reaches `main` with the next release.

Issue #285, opened 2026-09-11
fix(webview): createNonce's modulo reduction is biased — A–H are ~25% over-represented


eFAILution:
Follow-up from #275.

## Problem

`src/webview/webviewHtml.ts:26`:

```ts
nonce += NONCE_CHARS[byte % NONCE_CHARS.length];
```

`NONCE_CHARS` is 62 characters and `byte` ranges over 0–255. `256 % 62 == 8`, so indices 0–7 (`A`–`H`) are each reachable from 5 byte values while indices 8–61 are reachable from 4 — roughly 25% over-represented.

Not exploitable at 32 characters: even biased, the nonce carries far more entropy than an attacker could brute-force before the document is replaced. The reason to fix it is that the doc comment immediately above stakes a cryptographic claim —

> `crypto.getRandomValues` is available in the extension host runtime and is preferred over `Math.random` for a value that gates script execution.

— and a biased modulo reduction is the textbook mistake that claim is supposed to rule out. Someone reading this as the reference pattern for the other five webview documents will copy it.

## Suggested fix

Either:

- rejection-sample: discard bytes `>= 248` (the largest multiple of 62 below 256) and redraw, or
- use an alphabet whose size divides 256 — hex (16) or base64url (64). Both are valid CSP nonce characters, and base64url at 32 characters keeps the same entropy budget with no rejection loop.

The second is less code and has no loop to reason about.

eFAILution:
Closed by #305, merged to `beta` as b236672. Reaches `main` with the next release.

PR #305, opened 2026-09-21
chore(webview): declare asset resource roots and unit-test the CSP helpers

X-Guardian:
## Summary
- Second step of the webview asset work. Nothing consumes webview assets outside the loading view yet, so this PR is deliberately inert: it clears the way for the extractions that follow and adds the test coverage they will rely on.
- Fixes a latent blocker: the detached details panel declares `localResourceRoots: []`, which permits no local files at all. The moment that panel's CSS or scripts move to files they would 404 — on that path only, while the identical document rendered from the Component Browser looked correct.
- Splits the `vscode`-free half of `webviewHtml.ts` into `src/webview/csp.ts` so the nonce and CSP helpers can be unit-tested, and closes two findings from the #275 review while writing those tests.
- Link related issue(s): #288

## Change Type
- [ ] feat
- [x] fix
- [x] refactor
- [ ] docs
- [ ] test
- [x] chore (lint config)

### User-facing impact
- None. The only document currently using an external asset is the loading view, which is unaffected.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (webview plumbing only)

### Affected areas
- [x] Component Browser (panel options only)
- [x] Hover provider (the detached details panel's options)
- [ ] Completion provider
- [ ] Validation provider
- [ ] Docs only

## What changed by bucket

| Bucket | Files | Approach |
|---|---|---|
| Resource roots | `src/extension.ts`, `src/providers/componentBrowserProvider.ts`, `src/webview/webviewHtml.ts` | New `assetRoots(extensionUri)` beside `assetUri`, which already owns the `out/webview` layout. All three panels use it: the detached panel replaces `localResourceRoots: []`, the browser-path details panel gains an explicit declaration instead of relying on the default, and the Component Browser drops its hand-written path. |
| Testable CSP helpers | `src/webview/csp.ts`, `src/webview/webviewHtml.ts` | `createNonce` and `cspMetaTag` move to a `vscode`-free module, with `cspMetaTag(cspSource, nonce)` taking the origin as a string. `webviewHtml.ts` keeps `assetUri`/`assetRoots` and re-exports the pair, so callers still have one import. |
| Nonce distribution | `src/webview/csp.ts` | The nonce alphabet is 64 characters, so indexing by a random byte draws uniformly. At 62 the first two characters were over-represented by ~25%. |
| Asset path guard | `src/webview/webviewHtml.ts` | `assetUri` rejects empty paths and `.`/`..` segments, so a computed path cannot resolve outside the root its panel declares. |
| Sourcemap delivery | `esbuild.js` | Webview assets get an inline sourcemap in development. A sibling `.map` file makes the webview fetch it, which the document's `default-src 'none'` CSP blocks and reports — noise in the console that PRs 3-6 will be watching for real violations. Production builds emit no map either way. |
| Client lint | `eslint.config.js` | A block scoped to `src/webview/client/**/*.ts` with `globals.browser` plus `acquireVsCodeApi`. Scoped rather than applied to all TypeScript so extension-host code referencing `document` or `window` is still flagged. |
| Test coverage | `tests/unit/csp.test.ts` | Seven cases over the nonce (length, alphabet, uniqueness, distribution) and the CSP tag (`default-src 'none'`, nonce-gated scripts, no `'unsafe-inline'` on either directive). |

### Local checks
- [x] `npm run compile`
- [x] `npm test` — 438 passing (+7 new); `npm run lint` clean

### Manual test notes
- Opened the Component Browser and both details panels; the loading view still renders its spinner from `styles/loading.css`.

## Breaking Changes
- [x] No breaking changes

## Risk and Rollback
- Main risks: **low**. Declaring a resource root only widens what a panel may load, and the one document using an asset today already worked. The `assetUri` guard could in principle reject a path that previously resolved, but its only caller passes the literal `styles/loading.css`.

## Release Notes Draft
- Internal: webview panels now declare the resource root their assets are served from; no user-facing change.

## Checklist


eFAILution (approved):
Approving. Closes #284, #285 and #286 exactly as they were written, and picks up a real bug on the way.

The empty `localResourceRoots: []` on the detached panel is the one worth calling out — it permits no local files at all, so the identical document rendered from the Component Browser would have kept working while the hover path 404'd, which is the shape of bug that gets found late and blamed on the wrong change.

Verified:
- `createNonce` alphabet is 64 characters, so `byte % 64` divides 256 evenly and the bias is gone. The distribution test is the good kind — 2000 nonces, every character required to appear, max under 1.15x expected. It would fail on the old 62-character alphabet, which is what makes it worth having.
- `assetUri` throws on empty, `.` and `..`, and `ASSET_ROOT` is now the single place the `out/webview` layout is written down, shared with `assetRoots`.
- All three panels declare `assetRoots(...)`: the detached one, the browser-path details panel, and the Component Browser.

438 passing, lint and types clean, CI green. Scoping the browser globals block to `src/webview/client/**` rather than all TypeScript is the right call — extension-host code touching `document` still gets flagged.

PR #308, opened 2026-09-22
refactor(webview): serve the no-sources and error views from linted external assets

X-Guardian:
## Summary
- Third step of the webview asset work, and the first to put a document with scripts under a Content-Security-Policy. Covers the three smallest views — no-sources guidance, the per-source error list, and the single fatal error — as the lowest-risk test of the pattern the remaining two builders will copy.
- Their CSS and JS move out of TypeScript template literals into linted files under `src/webview/`, and their nine inline `onclick` attributes become delegated listeners, which a nonce CSP requires.
- Also removes two hardcoded placeholder components the cache injected when no sources were configured. They resolved to nothing, and their presence is what kept the no-sources guidance from ever rendering — so the view this PR extracts had no way to be reached, let alone verified.
- `src/providers/componentBrowserProvider.ts` loses 185 lines.
- Link related issue(s): #288

## Change Type
- [ ] feat
- [x] fix
- [x] refactor
- [ ] docs
- [ ] test
- [ ] chore

### User-facing impact
- **A user with no sources configured now sees the setup guidance**, in place of a `Local` source offering two 'demo' components that could not be inserted usefully.
- Otherwise none intended: the three views render and behave as before; only how their CSS and JS are delivered changed.

### GitLab scope
- [ ] gitlab.com
- [ ] self-managed GitLab
- [x] both (webview rendering only)

### Affected areas
- [x] Component Browser (the no-sources and error views)
- [ ] Hover provider
- [ ] Completion provider
- [ ] Validation provider
- [ ] Docs only

## What changed by bucket

| Bucket | Files | Approach |
|---|---|---|
| Stylesheets | `src/webview/styles/noSources.css`, `src/webview/styles/errors.css` | Lifted verbatim. The two error views share one stylesheet: `getErrorHtml`'s rules are a strict subset of `getErrorsHtml`'s apart from `.error`, so a second file would have been near-duplicate. |
| Client scripts | `src/webview/client/noSources.ts`, `src/webview/client/errors.ts` | The two error views' scripts were byte-identical, so they load one module. Handlers are delegated from the document on `data-action`, with `toggleDetails` reading `data-details-id`. |
| Client typecheck | `tsconfig.client.json`, `tsconfig.json`, `tsconfig.tests.json`, `package.json` | Client scripts get their own project, extending the root config and adding `@types/vscode-webview` for `acquireVsCodeApi` — a global that exists only in the renderer. Scoping it this way keeps extension-host code that references it failing to compile, which a `.d.ts` under `src/` would not: `declare global` there applies project-wide. `check-types` runs the new project, and both other configs exclude the directory. |
| Builders | `src/providers/componentBrowserProvider.ts` | All three take a `vscode.Webview`, emit a CSP meta tag, and reference their assets through `assetUri`. Their `<script>` carries the nonce. |
| Visibility without `style` | `src/webview/styles/errors.css`, `src/webview/client/errors.ts` | The raw-error blocks started hidden via `style="display: none;"`, which a nonce CSP blocks as surely as a `<style>` block. Initial state is now an `is-hidden` class and the toggle flips it with `classList`. |
| Stylelint config | `stylelint.config.mjs` | `custom-property-pattern` widened to admit VS Code's theme variables — `--vscode-textLink-foreground` and friends are camelCase after the component segment, and are read rather than declared here. `loading.css` happened to use only kebab-case names, so this surfaced now. |
| Deprecated property | `src/webview/styles/errors.css` | `word-break: break-word` → `overflow-wrap: break-word`, flagged by stylelint once the file was lintable. |
| Asset discovery | `esbuild.js` | Declaration files are excluded from the client directory scan, so a `.d.ts` there is not bundled into a stray `.d.js` asset. |
| Unreachable view | `src/services/cache/componentCacheManager.ts`, `src/constants/cache.ts` | With no sources configured the cache injected two hardcoded components, `deploy-component` and `test-component`, under a `Local` source. That made `getNoSourcesHtml` unreachable: it renders only when there are no sources **and** no components, and the fallback fired on exactly the first condition. Removed, along with the now-unused `SOURCE_LOCAL`. |
| Test coverage | `tests/unit/webviewAssets.test.ts` | Seven assertions over the build output: each asset is emitted and non-empty, each script is a self-contained IIFE with no leaked `export`/`import`, no declaration file is emitted, and no stylesheet references a sibling sourcemap. |

### On testing the builders
The plan for this step called for a guard asserting no builder's output contains an inline handler. That is not reachable: the builders are private methods on a class whose module imports `vscode`, so the unit suite cannot load it, and the extension-host project is rooted at `tests/extension-host` and cannot import `src/`. The tests here assert on the built assets instead — what the webview actually loads, and where a bundling or escaping fault would appear. Absence of inline handlers was checked by hand across the three builders.

### Local checks
- [x] `npm run compile`
- [x] `npm test` — 445 passing (+7 new); `npm run lint` clean (eslint and stylelint)

### Manual test notes
- Confirmed the build emits exactly the five expected assets, and that both client scripts parse as valid IIFEs.
- Checked in the Extension Development Host, with the webview console open for CSP violations:
  - **No sources** — *confirmed*: with `gitlabComponentHelper.componentSources` cleared and the cache reset, the guidance renders fully styled from the external stylesheet and **Open Settings** opens the Settings editor, so the delegated listener replaces the removed `onclick` correctly.
  - **Error list** — *confirmed*: with every source pointed at a missing project, the view renders styled from the external stylesheet and **Show details** reveals the raw error, flipping its label to **Hide details**. That covers the delegated `data-action` routing, `data-details-id` lookup, and the `is-hidden` class replacing the inline `style` attribute. **Try Again** and **Open Settings** both post correctly. **Update Token** is correctly absent for a 404, which is not an auth failure, so that branch is unexercised.
  - **Fatal error**: not exercised. It is the catch-all for an exception thrown during load, so it cannot be triggered deliberately without editing the code. It shares `errors.css` and `errors.ts` with the error list and has the same control set, so the verification above covers the same code paths.

## Breaking Changes
- [x] No breaking changes

## Risk and Rollback
- Main risks: **low-medium**. These views are only seen when something is already wrong, so a fault here lands on a user who is mid-problem — and the fatal-error view is the catch-all for an exception during load. A missed handler is a dead button rather than a visible error, which is why every control is listed above rather than summarised.

## Release Notes Draft
- Fix: with no component sources configured, the Component Browser now shows the setup guidance instead of two placeholder components that could not be resolved.
- Internal: the no-sources and error views load their CSS and JS from linted external files under a Content-Security-Policy.

## Checklist


eFAILution (changes requested):
The extraction itself is right, and the cache finding buried in the middle of it is the most valuable thing here — `getNoSourcesHtml` renders only when there are no sources **and** no components, and the fallback fired on exactly the first condition, so the view could never appear. Noticing that while extracting it, rather than extracting a view nobody could reach, is the part that took attention.

Verified across the three builders: all use `data-action` delegation, no `onclick`, no inline `style=`, no `<style>`. Each emits `cspMetaTag(webview.cspSource, nonce)`, a `<link>` to its stylesheet and a `<script nonce="…">`. The build emits exactly the five expected assets. 445 passing, lint clean, CI green.

One real bug, one lint rule that overshoots, and a loose end.

### Blocking

**`getErrorHtml` sets `class` twice, so the raw error renders expanded.**

`componentBrowserProvider.ts:2500`:

```html
<pre class="error-raw" id="error-raw" class="is-hidden">
```

HTML5 keeps the first `class` and discards later duplicates, so the element lands as `class="error-raw"` and `is-hidden` never applies. Compare `:2330`, which gets it right: `class="error-raw is-hidden"`.

The consequence isn't only the initial state — it inverts the toggle for the life of the document:

| | block | button |
|---|---|---|
| on open | **visible** | "Show details" |
| first click | hidden | "Show details" |
| second click | visible | "Hide details" |

So a fatal auth error opens with its raw text already expanded, and the button then contradicts what the user is looking at. That's a regression from the `style="display: none;"` it replaced, not just a cosmetic slip.

Nothing in the toolchain can catch it: it's inside a template literal, so `tsc`, eslint and stylelint all see a string. And your own notes name this as the one view not exercised — "it cannot be triggered deliberately without editing the code". That's fair, and it's also exactly where this landed. The `isAuth` branch narrows it further, which is why the error-list verification didn't surface it.

One character: `class="error-raw is-hidden" id="error-raw"`.

### Worth fixing before merge

**The `custom-property-pattern` override rejects any property the project declares itself.**

Your diagnosis is right — I confirmed stylelint flags these on `var()` reads, not just declarations, so the rule genuinely does fire on ten lines across the two stylesheets. But the replacement requires *every* custom property name to start with `vscode-`:

```
'custom-property-pattern': '^vscode-[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$'
```

Dropped a `--spacing-sm: 8px` into a probe stylesheet:

```
2:3   ✖  Expected "--spacing-sm" to match pattern "^vscode-[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$"
6:16  ✖  Expected "--spacing-sm" to match pattern "^vscode-[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$"
```

That's a normal thing to want as steps 4–6 extract bigger stylesheets, and the failure would arrive with a message that reads like a kebab-case complaint about a name that *is* kebab-case. The kebab-case discipline the standard config was providing is also gone for our own properties.

Admitting the prefix as an alternative keeps both:

```js
'custom-property-pattern': '^(vscode-[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*|[a-z][a-z0-9]*(-[a-z0-9]+)*)$',
```

Clean on both the probe and the real stylesheets.

### Loose end

**The same two placeholder components still exist in `componentService.ts`.**

`getLocalComponents()` at `:208` returns a byte-equivalent `deploy-component` / `test-component` pair, reached through `getComponents()` at `:82`. It doesn't affect this PR — nothing calls either (`grep` finds only the internal `:82` call; every `getComponents()` call site in `src/` is on the *cache manager*), so the view you unblocked stays unblocked. But someone removing a fake-component fallback will want to know there are two, and the surviving copy is dead code that reads as live.

### Not blocking

`acquireVsCodeApi()` is called at module scope in both client scripts. VS Code allows it once per document, and each document loads exactly one script today, so this is correct now. It becomes a constraint the moment a document wants two — worth knowing before the Component Browser view gets split in a later step.

Happy to push all three fixes if you'd rather not round-trip.

eFAILution:
Pushed the first two fixes in `2365b29`. **Withdrawing the third finding — I had it wrong.**

### Correction: the `componentService.ts` placeholders are not dead

I said `getLocalComponents()` was unreachable because every `.getComponents()` call site in `src/` is on the cache manager. That was a bad read of my own grep. The chain is live:

```
completionProvider.ts:56  getComponentUnderCursor(document, position)
componentDetector.ts:144    detectComponent(document, …)
componentDetector.ts:122      componentService.getComponents()
componentService.ts:82          getLocalComponents()   ← the two mocks
```

`tsc` caught it the moment I tried the removal — `Property 'getComponents' does not exist on type 'ComponentService'` from `componentDetector.ts`. Reverted; nothing in that file is touched by this PR now.

Worth knowing what it does, since it isn't nothing and isn't much either: `detectComponent` matches `line.includes('component: ' + name)` against whatever `getComponents()` returns, and that is only ever `deploy-component` and `test-component`. So the path compiles, runs on every completion request, and can only match two names that don't exist. That is its own question, and not one to settle inside a webview-assets PR.

### The two fixes

**`getErrorHtml`'s duplicate `class`** — now `class="error-raw is-hidden" id="error-raw"`, matching `:2330`.

Plus a guard, since the reason this slipped through is that no tool could see it. `tests/unit/webviewBuilderMarkup.test.ts` reads `componentBrowserProvider.ts` as text and rejects any element carrying the same attribute twice — the file can't be imported, but it can be read. Confirmed load-bearing:

```
1) componentBrowserProvider markup
     no element carries the same attribute twice:
   AssertionError: duplicate attributes:
   class repeated in: <pre class="error-raw" id="error-raw" class="is-hidden">
```

It covers all six builders, so it's in place before steps 4–6 move the two big ones.

**The stylelint pattern** — kept as an alternative rather than the whole pattern:

```js
'custom-property-pattern': '^(vscode-[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*|[a-z][a-z0-9]*(-[a-z0-9]+)*)$',
```

Clean on the real stylesheets and on a `--spacing-sm: 8px` probe, which the previous form rejected.

446 passing, eslint and stylelint clean, both typecheck projects clean. (`tsconfig.client.json` needs `@types/vscode-webview`, which isn't in `beta`'s tree yet, so that one only resolves in CI — it's green there.)

eFAILution (approved):
Approving — clearing my own changes-requested, since the two findings it was blocking on are fixed in `2365b29` and the third I withdrew as wrong.

**What landed:** the duplicate `class` in `getErrorHtml` is gone, so the fatal-error view opens collapsed again and its toggle label tracks the state. The `custom-property-pattern` override now admits the `vscode-` prefix as an alternative rather than as the whole pattern, so a property this project declares for itself still passes.

**What I got wrong:** I called the `componentService.ts` placeholder pair dead code. It isn't — `completionProvider` reaches it through `getComponentUnderCursor` → `detectComponent` → `getComponents()`. `tsc` caught me the moment I tried the removal. That file is untouched.

**What's worth keeping from it:** that path runs on every completion request and can only ever match `deploy-component` or `test-component`, two names that don't exist. Not this PR's problem, but it's the same fake-component fallback you removed from the cache here, still live on a different path. Worth its own issue.

The new `webviewBuilderMarkup` guard covers all six builders, so it's in place before steps 4–6 move the two big ones. It reads the provider as text rather than importing it, which sidesteps the `vscode` constraint you hit when you wanted a no-inline-handlers assertion — the same trick would work for that check if you want it in the next step.

446 passing, eslint and stylelint clean, all three typecheck projects clean, CI green.

